PR #36, opened 2014-08-06
Allow Null Response if allow_null is passed

prat0318:
If `allow_null` is passed, the `null` values are allowed. If `Pet` has `name: None`, even then `client.pet.getPetById(petId=2).result(allow_null=True)` is acceptable. If the param is not provided, swagger-py will fail the validations.

### DIFF
--- a/docs/source/quickstart.rst
+++ b/docs/source/quickstart.rst
@@ -105,7 +105,11 @@ will result in error like so:
 
 .. code-block:: console
 
-        TypeError: id's value: I should be integer :( should be in types (<type 'long'>, <type 'int'>)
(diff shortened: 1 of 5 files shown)


zroadhouse-rmn:
Did this option get removed along the way somewhere? I don't see allow_null in the current codebase.

prat0318:
Yes, this was present in the old version (which now resides in swaggerpy branch) and is now removed in bravado. Similar functionality has now been implemented using [x-nullable. Check the link.](https://github.com/Yelp/bravado-core/issues/47).

zroadhouse-rmn:
Thank you!

PR #41, opened 2014-08-08
Force headers implementation - override header values if passed through request

### LINKED ISSUE #31: Add support for 'header' request parameter

prat0318:
Fixes #31.

### DIFF
--- a/swaggerpy/async_http_client.py
+++ b/swaggerpy/async_http_client.py
@@ -49,10 +49,12 @@ class AsynchronousHttpClient(http_client.HttpClient):
         # request_params has mandatory: method, url, params
         if not request_params.get('headers'):
             request_params['headers'] = self._headers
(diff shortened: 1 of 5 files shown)


dnephin:
We have a feature that requires header support. @prat0318 let me know if you're able to work on this or not.  They need something by next wednesday, so I can take this over if you're occupied with other things.

prat0318:
yeah, i can pick it up next.

prat0318:
So, on reading again, this PR looks no longer useful. This PR read `header` param from the request (and obey the spec) as well as another optional attribute `headers_forced` which will override any of the attributes provided in the request's `header` param (don't remember the use-case of it though). Separately specifying the header attributes has now been covered in @e4b888f . I will create a separate PR fetching the `header` param as per spec from the request.

prat0318:
Closing it. Follow up in #122

PR #84, opened 2014-12-27
Use py.test instead of nose

dnephin:
I've found py.test is a lot nicer than nose. It works as a drop-in replacement, and it will let us write py.test style tests instead of unittest.

Also:
*  replaces tabs with spaces in `tox.ini`
* remove `lint` target, it doesn't actually pass. We can add it back if that's actually something we want to run, when it works


bpicolo:
Mind doing a rebase for clean merge?

PR #86, opened 2014-12-27
Cleanup swagger client

### LINKED ISSUE #76: If you specify a base_url for resource loading, it ignores basePath in all the resources
Seems like strange behavior to me. Might be worth taking a look at again at some point.

Edit: Actually, just ignores it always I'm pretty sure, though comments suggest otherwise = |

### LINKED ISSUE #80: stringify_body special case for datetime needs some fixing up
See pull #79.

"I think this should handle date as well (since those are both supported types in 1.2), and the datetime format this is using is not correct. I believe it should be using date-time from http://xml2rfc.ietf.org/public/rfc/html/rfc3339.html#anchor14, which is explicitly called out in the 2.0 spec. "

### LINKED ISSUE #83: Remove raise_with support from SwaggerClient
In #44 and #45 we added a `raise_with` param to the client. I believe we should remove support for this param:

* it can be handled without any code in swagger-py (see below)
* it's an attribute that gets added to the http_client, but it not actually part of the class. This makes it mutable, and we would run into errors like those we addressed in #68

Instead we can provide this (either as part of swagger-py or in documentation):
```python
class ErrorHandler(object):

    def __init__(self, client, handler):
        self.client = client
        self.handler = handler

    def __getattr__(self, name):
        def operation(*args, **kwargs):
            try:
                return getattr(self.swagger_client, name)(*args, **kwargs)
            exception Exception as e:
                self.handler(name, e)

client = ErrorHandler(SwaggerClient(...), my_error_handler)
```

dnephin:
Previously the entire process of fetching the api docs, validating the api docs, and building `Resources`/`Operations` was done as part of the constructor of `SwaggerClient`.  Constructors should never do anything more than assignment to fields.  Complex constructors are difficult to test, and make re-use more difficult. 

This branch splits the constructor into two factory methods `from_url()`, and `from_resource_listing()`.  This allows consumers to create a `SwaggerClient` from just a list of resources, which also removes the branch logic from the constructor. This is still backwards compatible because the only documented interface is the `get_client()` call. 

Some other things:
* use `simplejson` if it's available (since it's much faster, and this is very slow)
* completely removed the need for `ClientProcessor` (yay less inheritance!). This subclass was only being used for the single method that it implemented. All other operations are already done by default by the `Loader`. That method was extracted to `append_name_to_api()`.
* change the kwarg for the client cache to be `ttl` instead of `timeout` (`timeout` is still supported for backwards compatibility), and some other renames around the cache
* cache clients based on all args + kwargs, not just the first argument. All the arguments change the behaviour of the client, so they all need to be part of the key. Uses `repr()` instead of `json.dumps()`.
* `SwaggerClient` and `Resource` now support `dir()` (and tab completion in ipython) without `setattr` (which was duplicating the `__getattr__`).

Resolves #80. Remove special case for `datetime` in `stringify_body`. This should be totally unnecessary. The `swagger_model` should already convert this datetime into the proper format.

Resolves #76. Previously `api_base_path` was being used for two different things: overriding the default url that was specified in the swagger api doc, and also as a base url for swagger api docs that use relative urls.  I've split this into two params. `api_base_path` is now only an override, `base_url` is now the base url to append to relatives `basePath` urls.

Resolves #83. `raise_with` is unnecessary, and makes the interface for `swagger-py` more verbose than it needs to be.

Sorry this is so big. Most of this had to be changed all at once, so I wasn't able to break it into smaller parts. I did pull out a couple isolated commits, and a lot of this is just simple replacement (json imports, and test changes to use the new factories).

### DIFF
--- a/swaggerpy/client.py
+++ b/swaggerpy/client.py
@@ -243,6 +243,7 @@ class Resource(object):
     def from_api_doc(cls, api_doc, http_client, base_path, url_base=None):
         """
         :param api_doc: api doc which defines this resource
+        :type  api_doc: :class:`dict`
         :param http_client: a :class:`swaggerpy.http_client.HttpClient`
         :param base_path: base url to perform api requests. Used to override
                 the path provided in the api spec
--- a/swaggerpy/client.py
+++ b/swaggerpy/client.py
@@ -57,7 +57,6 @@ import logging
 import os.path
 import time
 import urllib
-from collections import namedtuple
 from urlparse import urlparse
 
 from yelp_uri import urllib_utf8
@@ -65,7 +64,11 @@ from yelp_uri import urllib_utf8
 import swagger_type
 from swaggerpy.http_client import APP_JSON, SynchronousHttpClient
 from swaggerpy.response import HTTPFuture, post_receive
-from swaggerpy.swagger_model import create_model_type, Loader
+from swaggerpy.swagger_model import (
+    Loader,
+    create_model_type,
+    is_file_scheme_uri,
+)
 from swaggerpy.swagger_type import SwaggerTypeCheck
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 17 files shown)


swaggerpy/client.py
```diff
@@ -1,94 +1,131 @@
...
+    :type  ttl: int
     """
 
-    def __init__(self, swagger_client, timeout, timestamp=None):
-        self.swagger_client = swagger_client
-        self.timeout = timeout
+    def __init__(self, item, ttl, timestamp=None):
+        self.item = item
+        self.ttl = ttl
         self.timestamp = timestamp or time.time()
```
analogue: I don't see why we need timestamp as a parameter. It seems like it should be internal to the implementation and always set to time.time().

If this is for testing, we're doing it wrong

swaggerpy/client.py
```diff
@@ -1,94 +1,131 @@
...
 
     def is_stale(self, timestamp=None):
         """Checks if the instance has become stale
```
analogue: Ditto

swaggerpy/client.py
```diff
@@ -1,94 +1,131 @@
...
+        return key in self.cache and not self.cache[key].is_stale()
 
-        if (key not in self.cache or self.cache[key].is_stale()):
-            self.cache[key] = self.build_cached_client(
-                api_docs_url, *args, **kwargs
-            )
+    def __call__(self, *args, **kwargs):
+        # timeout is backwards compatible with 0.7
+        ttl = kwargs.pop('ttl', kwargs.pop('timeout', SWAGGER_SPEC_CACHE_TTL))
+        key = repr(args) + repr(kwargs)
```
analogue: I'm in favor of just breaking compatibility and having less crufty code around unless you're going to go through a formal deprecation process.

swaggerpy/client.py
```diff
@@ -173,203 +195,207 @@ def _construct_request(self, **kwargs):
...
+        log.debug(u"Building resource '%s'" % name)
+        self._name = name
+        self._operations = operations
 
-    def _set_models(self):
-        """Create namedtuple of model types created from 'api_declaration'
+    @classmethod
+    def from_api_doc(cls, api_doc, http_client, base_path, url_base=None):
+        """
+        :param api_doc: api doc which defines this resource
```
analogue: Expected type would be helpful

tests/client_test.py
```diff
@@ -43,53 +35,55 @@ def test_unrequired_param_not_added_to_request_when_none(self):
...
+class SwaggerClientCacheTest(unittest.TestCase):
     """Test the proxy wrapper of SwaggerClient
     """
 
     def setUp(self):
-        client.factory = None
+        client.cache = None
 
-    def test_is_stale_returns_true_after_timeout(self):
+    tearDown = setUp
```
analogue: why?

analogue:
Looks good to merge even though large refactors like this scare me and turn my brain to swiss cheese. Definitely understand why this was not able to be split up into smaller pieces.

swaggerpy/client.py
dnephin: Yup, I think we could be patching `time.time()` here. Going to leave for another branch though since it's not modified in this one.

tests/client_test.py
dnephin: These tests mutate the global cache. They were cleaning up before running, but any other test that happens to hit the cache could fail if the test ordering changes.

This way the global state is reset after the tests as well.

Technically this could only be in tearDown and not setUp, but I'm not sure if all other tests honor this, but I left it in both for now

Issue #87, opened 2014-12-29
Improve swaggerpy performance


dnephin:
We're seeing that building the client for large APIs (~30 api docs) can be very slow (900ms!).  This ticket is to track some possible improvements.

From the profile:
 1. making requests for documents is ~700ms. We should do these async and all at once
 2. `create_model_docstring()` is 15ms, which is significant, and is only going to be used in very rare situations. It would be great to find a way to avoid this time, maybe with an optional parameter to enable them?

dnephin:
@prat0318 what do you thing about the second point?  

We could pass in a parameter, but it's pretty deep in the callstack so we'd have to pass it around a lot.  When do you see this docstring being used?

Since the models don't exist in code, they wouldn't be used by developers. It's possible that it could be used to document the class using sphinxdoc, but I would think that the swagger docs would be a better option.

Could we remove the model docstrings entirely?

I'm seeing them take around 8% of the time for some larger api docs.

prat0318:
I was instead thinking to make the `create_model_docstring` **lazy** instead of optional or removal. It can be used to show api-docs model information when working on REPL. Like so,
```
In [12]: p = swagger_client.pet.models.Pet
In [13]: p?
Attributes:

category (Category)
status (str) : pet status in the store
name (str)
tags (list(Tag))
photoUrls (list(str))
id (long) : unique identifier for the pet
```
My proposed patch prat0318@454b8d6 (inspired by [SO post](http://stackoverflow.com/a/5191224/987896) ) now loads model docstring lazily.

Earlier (added a print statement inside `create_model_docstring`):
```
In [2]: swagger_client = client.get_client(
   ...:     "http://petstore.swagger.wordnik.com/api/api-docs")

create_model_docstring called!
create_model_docstring called!
create_model_docstring called!
create_model_docstring called!
create_model_docstring called!
```

With the patch:
```
In [2]: swagger_client = client.get_client(
   ...:     "http://petstore.swagger.wordnik.com/api/api-docs")
In [12]: p = swagger_client.pet.models.Pet

In [13]: p?
create_model_docstring called!
Type:            type
String form:     <class 'swaggerpy.swagger_model.Pet'>
File:            /Users/prat0318/projects/swagger-py/swaggerpy/swagger_model.py
Init definition: p(self, **kwargs)
Docstring:
Attributes:
...
```
The patch is crude though, would need some refactoring.

dnephin:
[...]

PR #89, opened 2015-01-08
Async api docs

dnephin:
Step 1 of #87

6672537: addresses a bug I found when I tried to pass an async client to `get_client()`. the sync client only requires a url (it basically passes everything to `requests`), where the async client required a bunch of kwargs.

de81678: Remove wait/cancel from the http_client interface. These are really operations on the eventual objects returned by the client, not the client itself. Changed api-doc loading to support async loading of all linked api documents. This way all the requests are done at once.

6530d67: crochet blew up when it received a unicode url, so I'm handling it here. I think this will probably be fixed when we use sprocket (or we can handle it in sprocket if it's not already).


swaggerpy/async_http_client.py
```diff
@@ -46,36 +46,21 @@ def start_request(self, request_params):
         """
         # request_params has mandatory: method, url, params, headers
         request_params = {
-            'method': str(request_params['method']),
+            'method': str(request_params.get('method', 'GET')),
```
prat0318: it looks unnecessary, as 'method' is a required field of operation. Validation done earlier would throw up.

dnephin: We can't assume this is always called from operation.  

The reason for this change is that we're calling it from the Loader, which doesn't necessarily have to pass a method. Even if it does now, it's not required by `HttpClient` so I don't think it should be required here.

swaggerpy/async_http_client.py
```diff
@@ -46,36 +46,21 @@ def start_request(self, request_params):
...
-            'uri': str(request_params['url'] + '?' + urllib_utf8.urlencode(
-                request_params['params'], True))
+            'headers': listify_headers(request_params.get('headers', {})),
+            'uri': '%s?%s' % (
+                request_params['url'],
+                urllib_utf8.urlencode(request_params.get('params', []), True))
         }
 
+        # crochet only supports bytes for the url
+        if isinstance(request_params['uri'], unicode):
```
prat0318: shouldn't the check be with `basestring` instead?

dnephin: Nope, str is bytes. If it's already str, we don't want to encode it.

swaggerpy/response.py
```diff
@@ -53,7 +52,7 @@ def cancel(self):
         """Try to cancel the API (meaningful for Asynchronous client)
         """
         self._cancelled = True
-        self._http_client.cancel(self._request)
+        self._request.cancel()
```
prat0318: for synchronous, `SynchronousEventual` should have also defined `cancel()` for the above call?

dnephin: Yup, I think you're right. I'll add cancel

swagger.py.iml
```diff
@@ -1,19 +0,0 @@
...
-      <excludeFolder url="file://$MODULE_DIR$/swaggerpy.egg-info" />
-    </content>
-    <orderEntry type="jdk" jdkName="Python 2.7.2 virtualenv at ~/virtualenv/swagger" jdkType="Python SDK" />
-    <orderEntry type="sourceFolder" forTests="false" />
-  </component>
-  <component name="PackageRequirementsSettings">
-    <option name="requirementsPath" value="" />
-  </component>
-</module>
-
```
dnephin: I got tired of this thing getting in the way of my tab completion.

No one is using this anymore, right?

analogue:
Fix-n-ship!

PR #90, opened 2015-01-09
Lazy load model docstring

dnephin:
Resolves the second point in #87

Based on https://github.com/prat0318/swagger-py/compare/Lazy_model_docstring

Added some unit tests for model building which replaced a couple of the integration tests.

I did some manual testing as well, to verify the ipython interactive behaviour works as expected, and that `create_model_docstring` is not in the cprofile for a regular non-interactive call.

### DIFF
--- a/swaggerpy/swagger_model.py
+++ b/swaggerpy/swagger_model.py
@@ -308,7 +308,11 @@ class docstring_property(object):
 
 
 def create_model_type(model):
-    """creates a dynamic model from the model data present in the json
+    """Create a dynamic class from the model data defined in the swagger spec.
+
+    The docstring for this class is dynamically generated because generating
+    the docstring is relatively expensive, and would only be used in rare
+    cases for interactive debugging in a REPL.
 
     :param model: Resource model :class:`dict` with keys `id` and `properties`
     :returns: dynamic type created with attributes, docstrings attached
--- a/swaggerpy/swagger_model.py
+++ b/swaggerpy/swagger_model.py
@@ -1,21 +1,13 @@
 # -*- coding: utf-8 -*-
-
-#
-# Copyright (c) 2013, Digium, Inc.
-# Copyright (c) 2014, Yelp, Inc.
-#
-
-"""Code for handling the base Swagger API model.
-"""
-
+from copy import copy
+from functools import partial
 import logging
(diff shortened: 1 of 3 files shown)


tests/resource_model_test.py
```diff
@@ -168,19 +161,6 @@ def test_nones_for_dates_on_model_types_creation(self):
         )
 
     @httpretty.activate
-    def test_success_on_model_types_instantiation(self):
```
prat0318: The test name might be misleading, but i think the test checks the `__eq__` method implementation of the model. Is the test redundant / unnecessary?

prat0318:
lgtm.

tests/resource_model_test.py
```diff
@@ -145,13 +145,6 @@ def register_urls(self):
     ################################################################
 
     @httpretty.activate
-    def test_success_on_model_types_creation(self):
```
prat0318: Not sure, if this test is redundant?

dnephin: The tests for `vars()`, `dir()` and types cover this I think

tests/resource_model_test.py
dnephin: Yup, this is covered by line 37 in the new test

swaggerpy/swagger_model.py
```diff
@@ -307,35 +299,35 @@ def load_json(resource_listing, http_client=None, processors=None):
...
-            swagger_type.get_swagger_types(props))
-    # Define a class variable to store all required fields
-    setattr(model_type, '_required', model.get('required'))
-    return model_type
+        __dir__=lambda self: props.keys(),
+        _flat_dict=lambda self: create_flat_dict(self),
+        _swagger_types=swagger_type.get_swagger_types(props),
+        _required=model.get('required'),
+    )
+    return type(name, (object,), methods)
```
analogue: I think you need to call out in simple terms why we're doing this in the first place and the net effect: performance reasons and lazy docstring generation in REPL

analogue:
Fix-n-ship!

swaggerpy/swagger_model.py
dnephin: Good call, will document this

PR #92, opened 2015-01-15
Rename to bravado

analogue:
`make test` successful


wting:
:heart: Looks good to me. We need to also migrate the repository name to Yelp/bravado as well.

jnb:
:+1: lgtm

dnephin:
lgtm. This is going to cause tons of merge conflicts on import lines for the few branches I have in flight, but I can probably deal with that by doing the sed myself head of the merge

analogue:
@dnephin I'm going to limit the use of `bravado` and swagger 2.0 spec support to the swagger2.0 branch. I see the master branch as continuing to support 1.2 only.  The reason being that I really don't want to waste cycles on maintaining both 1.2 and 2.0 support in the same codebase. That way, we can  make a clean break from (swagger-py, 1.2) to (bravado, 2.0)

README.rst
```diff
@@ -1,8 +1,8 @@
-.. image:: https://travis-ci.org/Yelp/swagger-py.png?branch=master
-  :target: https://travis-ci.org/Yelp/swagger-py?branch=master
+.. image:: https://travis-ci.org/Yelp/bravado.png?branch=master
+  :target: https://travis-ci.org/Yelp/bravado?branch=master
 
 
-Swagger-py
+Bravado
 ==========
```
bwbaugh: Very minor: might consider deleting the extra `=` header chars here and elsewhere now that we're using a shorter name.

dnephin:
@analogue I'm not sure I understand the plan

`swagger-py` right now is kind of two things: a client (makes http requests), and a translator from schema types to python types (which also includes validation, `swagger_model`, and `swagger_type`).

Any fixes or cleanup to the client portion of the package should really be merged into the bravado branch to make it easier to merge `bravado` back into master at the end.

I think we should remove all of the validation and type handling stuff, and move it into the swagger_schema_validator package. That way we can use it from the server side as well (pyramid_swagger). If we do this, I suspect that bravado won't need to know anything about the schema, and would work with either one. Any differences are handled by swagger_schema_validator package.

analogue:
@dnephin I don't disagree, but from the offset it seems like quite a large refactoring. We're juggling quite a few things here that all add to the overall complexity of getting Swagger 2.0 functionality: sync vs async http client + sprocket, swagger 1.2 vs 2.0 codebase, and different validation techniques in bravado vs pyramid swagger. I'll keep this in mind going forward.

PR #93, opened 2015-01-21
merge master into swagger-2.0 branch

dnephin:
It will be easier to keep these in sync with bug fixes if we do it periodically I think


dnephin:
Since all of this was already reviewed before merging into master, going to proceed with this merge.

PR #94, opened 2015-01-22
Refactor loader

### LINKED ISSUE #71: _request_options should be available for api-docs request
Need to be able to pass similar _request_options to api docs requests

dnephin:
The `Loader` was previously a class, it has been converted to the function `load_resource_listing`, which removes the processors parameter. With this change we're starting to see how this code and https://github.com/Yelp/swagger_spec_validator will be able to converge.

Also changed the interface to support all request_parameters for fetching api-docs instead of only supporting headers. 

Moved `SwaggerError` into the `exceptions` module.

Resolves #71


swaggerpy/swagger_model.py
```diff
@@ -132,116 +133,80 @@ def cancel(self):
...
+        request_options=None):
+    """Load a complete swagger api spec and return all schemas compiled
+    into a single dict.
+
+    :param url: url to the swagger spec (file or http)
+    :param http_client: a :class:`swaggerpy.http_client.HttpClient` for
+        performing the requests to fetch api documents.
+    :param base_url: optional url to use as the base url for api doc paths
+    :param request_options: mapping of additional fields to specify in
+        the http request to fetch resources.
```
dnephin: This is now very close to https://github.com/Yelp/swagger_spec_validator/blob/master/swagger_spec_validator/util.py#L12

This supports a few extra parameters, and supports async loading, but otherwise they are functionally pretty similar. They both load and validate the swagger api docs.

One big difference is the return value.  The `swagger_spec_validator` only raises an exception on failure, but does nothing on success. I think we should talk about having the `swagger_spec_validator` actually return a python representation of the api docs.  This functionality is going to be re-usable in both the client (bravado) and the server (pyramid_swagger).

dnephin:
We actually do need this for swagger 1.2.

Adds support for a timeout on fetching api-docs

dnephin:
cc @bwbaugh @analogue @prat0318

analogue:
lgtm

PR #95, opened 2015-01-23
Use swagger spec validator

analogue:
Unit tests are obviously failing because this is a mish-mash of 1.2 and 2.0 code.  This is just the first step of many...

Useful notes:
- SwaggerProcessor and its only existing subclass, ValidationProcessor have been removed since spec validation is now the responsiblity of swagger_spec_validator
- Replaced `resource_listing` variable name/parameter with `spec` in shift from 1.2 to 2.0 verbage
- Verified manually that the spec is actually being validated with swagger_spec_validator in `tests/swagger_model/load_file_test.py`


dnephin:
once the tests pass I think this looks good.

I'll try and rebase my branch from #94 on this one. I expect it'll conflict quite a bit since many of the changes are similar.

PR #98, opened 2015-01-24
Fix cache key and repr for http clients.

### LINKED ISSUE #96: Fix key for `get_client(...)` caching
The client caching seems to be effectively disabled due to the changes introduced in 46a68a433bf106f7f3af113907d2305938bc0243. More specifically, I think that the problem manifests itself in line 65:
```
key = repr(args) + repr(kwargs)
```
For example, when `get_client(...)` gets called with the kwarg `http_client=AsynchronousHttpClient()`, then the return value of `repr(kwargs)` will include the specific memory address of the AsynchronousHttpClient object:
'http_client': \<swaggerpy.async_http_client.AsynchronousHttpClient object at **0x397ed10**\>

Therefore, actually a **unique key** is stored in the cache dict for every `get_client(...)` function call.

dnephin:
Resolves #96

cc @HaFl, @bwbaugh

### DIFF
--- a/docs/source/quickstart.rst
+++ b/docs/source/quickstart.rst
@@ -120,7 +120,14 @@ Caching
 
 .. note::
 
-        Caching can totally be bypassed by using ``SwaggerClient()`` (discouraged).
+    Caching can totally be bypassed by using
+    :class:`swaggerpy.client.SwaggerClient()` directly.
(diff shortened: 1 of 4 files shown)


analogue:
lgtm

PR #99, opened 2015-01-24
Add pre commit hooks

dnephin:


.pre-commit-config.yaml
```diff
@@ -0,0 +1,22 @@
+
+- repo: git://github.com/pre-commit/pre-commit-hooks
+  sha: 5fe82b
+  hooks:
+    - id: autopep8-wrapper
+      args: ['-i', '--ignore=E309']
```
analogue: I kinda like E309 - Add missing blank line (after class declaration).

dnephin: I think this one doesn't work correctly with docstrings. A blank line makes sense when there is no docstring (just a function), but this was adding blank lines between the class declaration and the docstring, which looked very wrong.

analogue:
ship-it!

dnephin:
rebased on master

PR #101, opened 2015-01-26
Support all request options for api-docs

dnephin:
and restore async loading of api docs


dnephin:
This is #94 based on the swagger2.0 branch

dnephin:
I guess this isn't necessary with 2.0

dnephin:
oops, wrong one

PR #103, opened 2015-02-03
Port resource and ops

analogue:
This branch shows models/resources/operations working (by no means complete) until something needs to be done with the response. Good stopping point before the the delta gets too large.

```python
from bravado import client
c = client.get_client('http://petstore.swagger.wordnik.com/v2/swagger.json')
pet = c.pet
future = pet.getPetById(petId=5)
result = future.result()
<insert bug ugly stack trace trying to make sense of the json response>
```

Notable changes:
- bravado.mapping.spec.Spec introduced to maintain context/state for the specification processing 
- more breaking things up and moving to bravado.mapping - now has `operation` and `param`
- pay particular attn to `build_resources` in `resource.py` where the mapping from resource->operation is new in the 1.2 -> 2.0 migration
- and of course, new unit tests for things which got tested indirectly


bravado/client.py
```diff
@@ -350,78 +193,21 @@ def from_url(
         """
         log.debug(u"Loading from %s" % url)
         http_client = http_client or SynchronousHttpClient()
-
-        # TODO: better way to customize the request for api-docs, so we don't
-        # have to add new kwargs for everything
```
dnephin: This TODO is still relevant

bravado/client.py
```diff
@@ -350,78 +193,21 @@ def from_url(
...
 
     @classmethod
-    def from_spec(cls, spec, http_client=None, origin_url=None):
+    def from_spec(cls, spec_dict, origin_url=None, http_client=None):
         """
         Build a :class:`SwaggerClient` from swagger api docs
 
-        :param spec: a dict with a Swagger spec in json-like form
-        :param http_client: an HTTP client used to perform requests
-        :type  http_client: :class:`bravado.http_client.HttpClient`
```
dnephin: http_client is still a parameter

bravado/mapping/docstring.py
```diff
@@ -58,3 +58,65 @@ def create_model_docstring(props):
             docstring += ": " + props[prop]['description']
         docstring += '\n\t'
     return docstring
+
+
+def create_operation_docstring(json_):
+    """Builds Operation docstring from the json dict
+
+    :param json_: data to create docstring from
+    :type json_: dict
```
dnephin: json dict is pretty generic (and not really relevant to this function).

I had been changing these to "spec" or something like that.  Maybe "operation_definition" ?

dnephin:
It's a bit difficult to tell what was changed and what was moved but seems fine to me for another iteration.

Definitely +1 on the new testing

bravado/client.py
analogue: Restored

bravado/client.py
analogue: The `http_client` is now an attribute of the `bravado.mapping.spec.Spec` class so it is not retained by `SwaggerClient` directly.

dnephin: I see it as a parameter on this function, but the docstring for that parameter was removed

bravado/mapping/docstring.py
analogue: I started calling the "fragments" from the spec as <name>_dict but I think <name>_spec is more appropriate too. I'll try to make the change globally in the next pull request.

PR #104, opened 2015-02-12
Refactor parameter validation + serialization

analogue:
Sorry for the large pull request - it kind of grew and grew.

Context: Swagger has definitions, paths, parameters, responses, etc.  This review is mostly about parameters (ignore the branch name). 

A few notes about naming:
- ```*_spec``` : dict fragments from the swagger spec (except when talking about the spec itself -- swagger_spec which is a ```bravado.mapping.spec.Spec``` class
- ```*_dict```: tried to stop using this for anything related directly to the spec - only use for the request_dict atm

Notes:
- wrt parameters, primitive and array is implemented. Body and formData are TODOs
- this contains the docstring changes for operations in ```docstring.py```. ```help(petstore.pet.getPetById)``` now works
- anything that is ```$ref```able is stored under two keys for - one is the simple name and the other is the ```#/``` style absolute format. May do away with the simple name thing later.
- ```marshal.py``` is new and contains validation and serialization stuff for parameters only atm
- ```schema.py``` is new and contains Swagger Spec -> json schema conversion stuff so that we can delegate as much per request/response validation off to jsonsonschema
- I'd like to get rid of 1.2 legacy code ```validate_and_add_params_to_request```, ```add_param_to_request```, ```SwaggerTypeCheck``` and related classes/funcs because it is just doing too much stuff and is hard to follow. None of the new parameter stuff uses them.
- ```tests/petstore/*``` was a small diversion into hitting the 2.0 petstore API directly - most of the error cases don't work because of the server side impl.  Provided a good way to actually hit an API when I made changes. These tests are excluded from test discovery when doing a build
- $ref resolution is still a work in progress - operation parameters that $ref to ```#/parameters``` works
- I still don't like building dicts with overly verbose comprehensions, but I have a new reason not to like them -- stepping through them in the debugger is crap
- You know i'm a unit test nut! I was pretty disciplined but I'm sure there are things I missed...

Expect to couple follow up commits to clean things up and get flake8 passing.


bravado/http_client.py
```diff
@@ -16,6 +16,8 @@
 
 
 log = logging.getLogger(__name__)
+log.setLevel(logging.DEBUG)
```
prat0318: was `debug` level there for some testing?

bravado/http_client.py
```diff
@@ -153,6 +155,7 @@ class SynchronousHttpClient(HttpClient):
 
     def __init__(self):
         self.session = requests.Session()
+        #self.session.proxies = {'http':'http://localhost:9999'}
```
prat0318: remove it ? ^

bravado/mapping/docstring.py
```diff
@@ -60,6 +60,45 @@ def create_model_docstring(props):
...
+
+    :param operation: :class:`Operation`
+    :return: callable that executes the operation
+    """
+    def wrapper(**kwargs):
+        return operation(**kwargs)
+
+    wrapper.__doc__ = create_operation_docstring(operation)
+    wrapper.__name__ = str(operation.operation_id)
+    return wrapper
```
prat0318: The wrapper is to make the docstring loading lazy, correct?

bravado/mapping/docstring.py
```diff
@@ -82,58 +121,63 @@ def create_operation_docstring(operation):
         :type from_date: str
         :rtype: list
     """
+    print 'creating op docstring for %s' % operation.operation_id
```
prat0318: It should be either a log or appended to `s`

bravado/mapping/docstring.py
```diff
@@ -82,58 +121,63 @@ def create_operation_docstring(operation):
...
         s += "[{0}] {1}\n\n".format(operation.http_method.upper(), summary)
 
-    desc = op_dict.get('description')
+    desc = op_spec.get('description')
     if desc:
         s += "{0}\n\n".format(desc)
 
-    for param_dict in op_dict.get('parameters', []):
-        s += create_param_docstring(param_dict)
+    for param_spec in op_spec.get('parameters', []):
```
prat0318: Shouldn't we also add parameters listed for the path and common to all operations within. Not sure if they are already included when building this `operation` dict from the spec.

bravado/mapping/docstring.py
```diff
@@ -82,58 +121,63 @@ def create_operation_docstring(operation):
...
+        s += create_param_docstring(param_spec)
 
-    responses = op_dict.get('responses')
-    for http_status_code, response_dict in iter(sorted(responses.iteritems())):
-        response_desc = response_dict.get('description')
+    responses = op_spec.get('responses')
```
prat0318: It could lead to responses being `None` and later `None.iteritems()` throwing up. Though, `responses` should atleast contain one element, i would still add a default value `{}` along.

bravado/mapping/docstring.py
```diff
@@ -82,58 +121,63 @@ def create_operation_docstring(operation):
...
+        response_desc = response_spec.get('description')
         s += ':returns: {0}: {1}\n'.format(http_status_code, response_desc)
-        schema_dict = response_dict.get('schema')
-        if schema_dict:
+        schema_spec = response_spec.get('schema')
+        if schema_spec:
             s += ':rtype: {0}\n'.format(
-                swagger_type.get_swagger_type(schema_dict))
+                swagger_type.get_swagger_type(schema_spec))
     return s
```
prat0318: a TODO for reading the `examples` if present would be good.

bravado/mapping/docstring.py
```diff
@@ -82,58 +121,63 @@ def create_operation_docstring(operation):
...
+    for param_spec in op_spec.get('parameters', []):
+        s += create_param_docstring(param_spec)
 
-    responses = op_dict.get('responses')
-    for http_status_code, response_dict in iter(sorted(responses.iteritems())):
-        response_desc = response_dict.get('description')
+    responses = op_spec.get('responses')
+    for http_status_code, response_spec in iter(sorted(responses.iteritems())):
+        response_desc = response_spec.get('description')
         s += ':returns: {0}: {1}\n'.format(http_status_code, response_desc)
```
prat0318: Open question: should the 'default' response be shown separately? i think it is ok to put it with others.

bravado/mapping/docstring.py
```diff
@@ -82,58 +121,63 @@ def create_operation_docstring(operation):
...
         :param status: Status to be considered for filter
         :type status: string
     """
-    name = param_dict.get('name')
-    desc = param_dict.get('description', 'Document your spec, yo!')
-    default_value = param_dict.get('default')
-    location = param_dict.get('in')
+    name = param_spec.get('name')
+    desc = param_spec.get('description', 'Document your spec, yo!')
+    desc = desc if desc else 'Document your spec, yo!'
```
prat0318: Open: Is this needed? This means 'description' has deliberately been put as "".

bravado/mapping/marshal.py
```diff
@@ -0,0 +1,63 @@
...
+            value = [param.default]
+
+    if param.required and value is None:
+        raise SwaggerError('Parameter {0} cannot be null'.format(param.name))
+
+    jsonschema.validate(value, param.jsonschema)
+    return value
+
+
+def marshal_array(param, value, request):
```
prat0318: I was thinking if `marshal_array` and `marshal_primitive` can be clubbed together. `validate_array` can throw up if `param.location` turns out to be `path`.

bravado/mapping/marshal.py
```diff
@@ -0,0 +1,63 @@
...
+    if param.location == 'query':
+        request['params'][param.name] = value
+    elif param.location == 'header':
+        request['headers'][param.name] = value
+    elif param.location == 'formData':
+        raise NotImplementedError('TODO')
+    elif param.location == 'body':
+        raise NotImplementedError('TODO')
```
prat0318: If not clubbing, even this should raise an unknown location `else` error.

bravado/http_client.py
analogue: ack

bravado/http_client.py
analogue: ack

bravado/mapping/docstring.py
analogue: It is not lazy atm.  I'll add a TODO for that. The wrapper is to attach the docstring to a callable.

bravado/mapping/docstring.py
analogue: ack

bravado/mapping/docstring.py
analogue: Added TODO

bravado/mapping/docstring.py
analogue: ```responses``` is required. If not specified, ssv will catch it.  No need to handle a scenario that can't happen.

bravado/mapping/docstring.py
analogue: ack

bravado/swagger_type.py
```diff
@@ -26,6 +28,13 @@
     'File': file
 }
 
+SWAGGER20_PRIMITIVES = [
+    'integer',
+    'number',
+    'string',
+    'boolean',
+    'null'
```
prat0318: Is null a valid type?

bravado/mapping/docstring.py
analogue: Yes, this is intentional to discourage empty descriptions.

bravado/swagger_type.py
```diff
@@ -133,11 +151,18 @@ def is_array(type_):
...
+    """
+    return type_ == 'object'
+
+
 def is_complex(type_):
     """checks whether the swagger type is neither primitive nor array
     :rtype: boolean
     """
-    return not is_primitive(type_) and not is_array(type_)
+    return (not is_primitive(type_) and not is_array(type_)) or is_object(type_)
```
prat0318: curious if there would a case where type is not primitive and not array, and not even an object?

bravado/swagger_type.py
```diff
@@ -324,29 +353,44 @@ def _check_array_type(self):
...
+                "Type for {0} is expected to be a dict but is {1) instead".format(
+                    self.value, type(self.value)))
+
     def _check_complex_type(self):
         """Checks all the fields in the complex type are of proper type
+
         All the required fields are present and no extra field is present
         """
-        klass = self.models[self._type]
+        print type(self.value)
```
prat0318: remove print

bravado/response.py
```diff
@@ -123,6 +123,11 @@ def post_receive(response, type_, models, **kwargs):
     if kwargs.pop('raw_response', False):
         return response
 
+    # Response had data, but responses in spec didn't match it so type_ is
+    # not known
+    if type_ is None:
+        return response
```
prat0318: Should we log a warning for this?

bravado/mapping/marshal.py
analogue: I recollect something about the ```path``` not supporting arrays.  How would that look exactly?

bravado/swagger_type.py
prat0318: May be, it could be complex object, but how exactly is object and complex object different? I see two different create methods above.

bravado/mapping/marshal.py
analogue: What does "clubbing" mean? New term to me

bravado/mapping/schema.py
```diff
@@ -0,0 +1,82 @@
+from bravado.exception import SwaggerError
+from bravado import swagger_type
+
+
+JSONSCHEMA_PRIMITIVE_VALIDATIONS = {
+    'string': ('minLength', 'maxLength', 'pattern', 'format', 'enum', 'default'),
+    'integer': ('multipleOf', 'minimum', 'maximum', 'exclusiveMaximum', 'exclusiveMinimum', 'enum', 'default', 'format'),
```
prat0318: i think it will exceed 80. Also, is there really a need to pick these out separately for each type? Doesnt jsonschema just ignore the fields not applicable?

bravado/mapping/spec.py
```diff
@@ -38,17 +56,47 @@ def build(self):
         validator20.validate_spec(self.spec_dict)
```
prat0318: as we have the version differentiator in place in ssv, we can call the generic validating method.

bravado/mapping/spec.py
```diff
@@ -38,17 +56,47 @@ def build(self):
...
-        # TODO
-        # self.shared_parameters =
-        #   build_parameters(self.spec_dict.get('parameters', {})
+    def build_params(self):
+        params = {}
+        for ref_name, param_spec in self.spec_dict['parameters'].iteritems():
+            # Register under both 'plain' ref name and #/ style ref name
+            params[ref_name] = param_spec
+            params['#/parameters/{0}'.format(ref_name)] = param_spec
+        return params
```
prat0318: any specific reasons why `build_models` is placed separately

bravado/mapping/spec.py
```diff
@@ -38,17 +56,47 @@ def build(self):
...
+        :param object_spec: object schema
+        :return: dict
+        :rtype: dict
+        """
+        if len(object_spec) == 1 and '$ref' in object_spec:
+            # refs can be to
+            # - #/parameters
+            # - #/definitions
+            # - #/responses
+            #
```
prat0318: I am not sure if this is a correct way of resolving. We are narrowing the 'ref' reference either to be absolute from current file name or just the param name. But i think there are numerous other ways of stating a correct ref. May be, absolute path like 'File:///' or it could also be in another file. `RefResolver` in jsonschema is somewhat which can be looked at for this purpose.

bravado/mapping/marshal.py
prat0318: Yeah, i meant to say that a singal method can be used to marshal array and primitive. As primitive also has a `param.location == 'path'` condition, i would have a check in `validate_array` to raise if param's location is `path`.

bravado/mapping/marshal.py
prat0318: Oops i placed the comment at wrong line. It was meant to be for line no. 63, if not combining `marshal_primitive` and `marshal_array`, then `marshal_array` should also have a similar `else` condition.

bravado/mapping/operation.py
```diff
@@ -54,49 +93,84 @@ def operation_id(self):
...
+        """
+        Given the parameters passed to the operation invocation, validates and
+        marshals the parmameters into the request dict.
+
+        :type request: dict
+        :param op_kwargs: the kwargs passed to the operation invocation
+        :raises: TypeError on extra parameters or when a required parameter
+            is not supplied.
+        """
+        current_params = self.params.copy()
```
dnephin: Why copy? we shouldn't be mutating this anywhere

tests/mapping/operation/construct_params_test.py
```diff
@@ -0,0 +1,66 @@
...
+    test_with_mocks()
+
+
+def test_no_params(empty_swagger_spec, request_dict):
+
+    @patch('bravado.mapping.operation.marshal_param')
+    def test_with_mocks(mock_marshall_param):
+        op = Operation.from_spec(empty_swagger_spec, '/pet', 'get', {})
+        op.construct_params(request_dict, op_kwargs={})
+        assert 0 == mock_marshall_param.call_count
```
dnephin: s/marshall/marshal/

bravado/mapping/operation.py
dnephin: Oh, nevermind, it is being mutated by `pop()`.

bravado/mapping/marshal.py
```diff
@@ -0,0 +1,63 @@
...
+    jsonschema.validate(value, param.jsonschema)
+
+    # TODO: transform
+    return value
+
+
+def marshal_primitive(param, value, request):
+    if param.location == 'path':
+        token = u'{%s}' % param.name
+        request['url'] = request['url'].replace(token, urllib.quote(unicode(value)))
```
dnephin: I really want to say that this can be done with `string.format()`, but in order to make that work, this whole thing would have to change. Instead of doing param at a time, it would have to group everything by `param.location`, then validate every param separately, and finally build the request field all together at once with all param/values pairs for the same `param.location`.

This is something I wanted to do with the original swagger-py, but I suspect you aren't going to want to do that at this stage, so I guess we can leave it for now.

Another plus I see with this approach is that it makes this code immutable. We don't have to pass a request dict around and mutate it all over.

bravado/mapping/marshal.py
```diff
@@ -0,0 +1,63 @@
...
+        # Wrap if necessary - spec is weird about this
+        if type(param.default) == list:
+            value = param.default
+        else:
+            value = [param.default]
+
+    if param.required and value is None:
+        raise SwaggerError('Parameter {0} cannot be null'.format(param.name))
+
+    jsonschema.validate(value, param.jsonschema)
```
dnephin: Is this for validating each item in the array?  I believe swagger-py supports types that aren't supported by jsonschema, no?

I also worry about using jsonschema on every request. It's known to be terribly slow. Great for static validation, but painful for live requests.

bravado/mapping/operation.py
```diff
@@ -1,33 +1,72 @@
...
+
+        :return: list of dicts representing parameter specs
+        """
+        specs = []
+        for param_spec in self.operation_spec.get('parameters', []):
+            resolved_param_spec = self.swagger_spec.resolve(param_spec)
+            specs.append(resolved_param_spec)
+        return specs
+
+    def build_params(self):
```
dnephin: -1 on mutating this class after it's constructed.

Can this be done as a module level function and just pass in `params` to the constructor?  I guess this applies to `resolved_param_specs` as well. Neither of these are referenced on the instance itself after it's created.

That way the logic for building an `Operation` is clearly separated from the functionality of the operation.

bravado/mapping/operation.py
```diff
@@ -54,49 +93,84 @@ def operation_id(self):
...
+                    log.warn("Op {0} was successful by didn't match any responses".format(self.operation_id))
+                else:
+                    raise SwaggerError("Response doesn't match any expected responses: {0}".format(response))
+
+            response_dict = response.json()
+
+            if response_spec and 'schema' in response_spec:
+                swagger_type_ = swagger_type.get_swagger_type(response_spec['schema'])
+            else:
+                swagger_type_ = None
```
dnephin: line 151 to here should probably be extracted as a new function `get_swagger_response_type()` or something like that since this function is starting to get pretty big.

bravado/mapping/operation.py
```diff
@@ -54,49 +93,84 @@ def operation_id(self):
...
+
+            log.debug('response_dict = %s' % response_dict)
+            log.debug('response_spec = %s' % response_spec)
+            log.debug('swagger_type  = %s' % swagger_type_)
```
dnephin: This is potentially expensive even when the debugging is turned off (having to `str` all of these).

You can use the form:

```python
log.debug('response_dict = %s', response_dict)
```

That way the object `__str__` is only called if the thing is actually going to be logged somewhere.

bravado/mapping/param.py
```diff
@@ -1,32 +1,35 @@
...
+from bravado.mapping.schema import to_primitive_schema, to_array_schema
+from bravado.swagger_type import SwaggerTypeCheck, SWAGGER20_PRIMITIVES
 
-# TODO: Needs to be thought through some more instead of just carrying over
-#       the impl from 1.2
 
-def validate_and_add_params_to_request(spec, param_dict, value, request):
-    """Validates if a required param_dict is given and wraps 'add_param_to_req'
+# TODO: remove
+def validate_and_add_params_to_request(spec, param_spec, value, request):
```
dnephin: yup, looks unused, going to skip this section, agree with remove

bravado/mapping/schema.py
```diff
@@ -0,0 +1,82 @@
...
+    # identify the validation keys in the swager spec that are supported by
+    # jsonschema for the given primitive type
+    primitive_type = primitive_spec['type']
+    transferable_keys = ['name', 'type', 'description']
+    transferable_keys += JSONSCHEMA_PRIMITIVE_VALIDATIONS[primitive_type]
+    schema = {}
+    for key, value in primitive_spec.iteritems():
+        if key in transferable_keys:
+            schema[key] = value
+    return schema
```
dnephin: swagger supports more types than json, so this isn't going to cover many of the cases. On top of that I think this is more work (and a lot slower), than just checking the types directly.

We're going to want to be able to support user-defined formats as well, which won't work with this model, but should be a simple extension if we have our own `type` -> `validator` mapping already.

bravado/mapping/spec.py
```diff
@@ -38,17 +56,47 @@ def build(self):
         validator20.validate_spec(self.spec_dict)
         self.api_url = build_api_serving_url(self.spec_dict, self.origin_url)
         self.definitions = build_models(self.spec_dict['definitions'])
+        self.params = self.build_params()
+        self.responses = self.build_responses()
+        self.resources = build_resources(self)
```
dnephin: Same comment about mutable objects.

I think these should all be part of either a separate factory class, or just plain functions on the module which can be used as part of a factory classmethod `from_resource_spec()` or something like that.

bravado/mapping/spec.py
```diff
@@ -38,17 +56,47 @@ def build(self):
...
 
-        self.resources = build_resources(self)
+    def resolve(self, object_spec):
+        """Given an object specification, resolve the $ref if it is a $ref
 
+        :param object_spec: object schema
+        :return: dict
+        :rtype: dict
+        """
+        if len(object_spec) == 1 and '$ref' in object_spec:
```
dnephin: In jsonschema you can have a `$ref` and a `description`, from https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#schema-object- it seems like this is also true for swagger spec.

So this check of len == 1 is going to miss anything with extra fields, right?

bravado/mapping/spec.py
```diff
@@ -38,17 +56,47 @@ def build(self):
...
+            # refs can be to
+            # - #/parameters
+            # - #/definitions
+            # - #/responses
+            #
+            # TODO: consider tightening up the criteria by having the
+            #       caller specify in which context the ref is valid.
+            #       Search all for now
+            ref_name = object_spec['$ref']
```
dnephin: +1 for this TODO

bravado/mapping/marshal.py
dnephin: The extra "types" are the "formats" I guess, so this (and the above) isn't going to cover dates, bytes, or any of the extensions for user specified formats.

bravado/mapping/docstring.py
analogue: Added TODO

bravado/mapping/marshal.py
analogue: Re-wrote to be non-parameter specific.

bravado/mapping/marshal.py
analogue: Re-wrote to be non-parameter specific

bravado/swagger_type.py
analogue: TBD if this is of any use in swagger. Added TODO

bravado/swagger_type.py
analogue: ```is_complex()``` is no longer used in the refactored code. It just has diffs as I was trying to contort the 1.2 implementation to do things it wasn't meant to do.

bravado/swagger_type.py
analogue: ack

bravado/response.py
analogue: Will re-visit in subsequent pull request. Just jammed this in to get petstore responses working temporarily

bravado/mapping/spec.py
analogue: The generic version insists on downloading the spec itself.

bravado/mapping/spec.py
analogue: ```build_params(...)``` is no more

bravado/mapping/spec.py
analogue: All ref-resolving code is delegated to ```jsonref``` package.  See ```Spec.from_dic(...)``` for the new code flow.

tests/mapping/operation/construct_params_test.py
analogue: ack

bravado/mapping/marshal.py
analogue: Added a skeleton for the ```format``` stuff in ```formatter.py```.  Work in progress..

bravado/mapping/operation.py
analogue: This is gone because I'm now using ```jsonref```

bravado/mapping/operation.py
analogue: I'm going to do this in the next pull-request since this is a ```response``` thing and I'm trying to wrap up ```parameter``` stuff in this PR. I have a TODO in there to rewrite/refactor.

bravado/mapping/operation.py
analogue: Good to know

bravado/mapping/schema.py
analogue: See ```formatter.py``` a skeleton of the ```format``` support.

bravado/mapping/spec.py
analogue: ```build_params(...)``` is done.  Will refactor others in another pull request.

bravado/mapping/spec.py
analogue: Replaced by using ```jsonref``` library.

bravado/mapping/spec.py
analogue: Replaced by using ```jsonref```` library.

analogue:
Ok, I'm going to get flake8 stuff cleaned up and close scope for this PR.

bravado/mapping/formatter.py
```diff
@@ -0,0 +1,65 @@
...
+def date_from_wire(value):
+    # TODO: None handling
+    # TODO: Fix me - this is obviously not correct
+    value = datetime.date(value)
+    return value
+
+
+_formatters = {
+    'date': (date_to_wire, date_from_wire, 'Converts a date to/from a string'),
+    'int64': (lambda i: long(i), lambda i: long(i),
```
prat0318: I am not sure if both the to_wire and from_wire for int64 should be long(i), shouldn't there be str() for to_wire? And may be a TODO: for other types?

requirements.txt
```diff
@@ -1,3 +1,4 @@
 # TODO(spatel): Remove once integrated with a packaged version of swagger_spec_validator
 #-e /nail/home/spatel/git/swagger_spec_validator_analogue
 -e git://github.com/Yelp/swagger_spec_validator.git@#egg=swagger-spec-validator
+jsonref
```
prat0318: we can place it together with others in `setup.py`'s install_requires?

analogue: good catch. fixed

bravado/mapping/model.py
```diff
@@ -144,3 +154,49 @@ def create_flat_dict(model):
...
+        # only tag objects. Not all #/definitions map to a Model type - can
+        # be primitive or array, for example
+        if model_type == 'object':
+            model_spec[MODEL_MARKER] = model_name
+
+
+def fix_malformed_model_refs(spec):
+    """jsonref doesn't understand  { $ref: Category } so just fix it up to
+    { $ref: #/definitions/Category } when the ref name matches a #/definitions
+    name. Yes, this is hacky!
```
prat0318: This hack also needs to be put for dereferencing parameters and responses as well?

bravado/mapping/formatter.py
analogue: yeah, i just mushed this in there to get a test passing. will add TODO.

bravado/mapping/model.py
analogue: Will get to that bridge when it comes.

dnephin: I think we're better off doing this in a fork on jsonref (and see about getting it upstream) than inline in bravado.  

Maybe a TODO for now?

bravado/mapping/operation.py
```diff
@@ -1,33 +1,65 @@
...
+        """
+        self.params = {}
+        op_param_specs = self.op_spec.get('parameters', [])
+        path_specs = self.swagger_spec.spec_dict['paths'][self.path_name]
+        path_param_specs = path_specs.get('parameters', [])
+        param_specs = op_param_specs + path_param_specs
+
+        for param_spec in param_specs:
+            param = Param(self.swagger_spec, param_spec)
+            self.params[param.name] = param
```
dnephin: TODO to make this a factory?

bravado/mapping/model.py
analogue: ack

bravado/mapping/operation.py
analogue: ack

PR #105, opened 2015-02-22
Unmarshal responses

analogue:
Refactor response handling for the synchronous Requests lib client.

- Includes swagger responses of type primitive, list, object, and Model
- Response adapter introduced to make future http client libs more easy to integrate (fido)
- Unit tests as usual
- Removed dead code


bravado/mapping/unmarshal.py
```diff
@@ -0,0 +1,143 @@
...
+    - validate that the value conforms to the schema_object_spec
+    - transform the value according to 'format' if available
+    - return the value in a form suitable for use. e.g. conversion to a Model
+      type.
+
+    :type swagger_spec: :class:`bravado.mapping.spec.Spec`
+    :type schema_object_spec: dict
+    :type value: int, float, long, string, unicode, boolean, list, dict
+    :return: unmarshaled value
+    :rtype: int, float, long, string, unicode, boolean, list, dict, Model type
```
prat0318: DateTime?

bravado/mapping/response.py
```diff
@@ -0,0 +1,30 @@
+class ResponseLike(object):
+    """
+    Define a common interface for bravado to interface with client side
+    response objects from various 3rd party libraries.
+
+    Subclasses are responsible for providing attrs for __required_attrs__.
+    """
+    __required_attrs__ = [
+        'status_code',  # int
+    ]
```
dnephin: isn't `json()` part of this as well? seems to be used

bravado/mapping/response.py
```diff
@@ -0,0 +1,30 @@
+class ResponseLike(object):
+    """
+    Define a common interface for bravado to interface with client side
+    response objects from various 3rd party libraries.
+
+    Subclasses are responsible for providing attrs for __required_attrs__.
+    """
```
dnephin: So my understanding of this class is that its the `Response` interface that bravado expects, and can also be used as a  subclass to provide more friendly error messages to developers trying to implement this interface.

I usually prefer to avoid inheritance and let them debug the attribute error, but I'm ok with this too.

dnephin:
lgtm

bravado/mapping/unmarshal.py
```diff
@@ -0,0 +1,143 @@
...
+def unmarshal_model(swagger_spec, model_spec, model_value):
+    """Unmarshal a dict into a Model instance.
+
+    :type swagger_spec: :class:`bravado.mapping.spec.Spec`
+    :type model_spec: dict or jsonref.JsonRef
+    :type model_value: dict
+    :rtype: Model instance
+    :raises: TypeError
+    """
+    model_name = model_spec[MODEL_MARKER]
```
prat0318: is this `x-model` stated in 2.0 spec? or an assumption.

dnephin: I believe `x-` are use extensions in swagger 2.0 , I'm actually not sure how that relates to `x-model` here.

Is this something that we're going to require for swagger-py support?

bravado/mapping/unmarshal.py
```diff
@@ -0,0 +1,143 @@
...
+    """
+    obj_type = schema_object_spec['type']
+
+    if obj_type in SWAGGER20_PRIMITIVES:
+        return unmarshal_primitive(schema_object_spec, value)
+
+    elif obj_type == 'array':
+        return unmarshal_array(swagger_spec, schema_object_spec, value)
+
+    elif is_model(schema_object_spec):
```
prat0318: Is the difference between a model and an object is that model is defined as a `ref` in #/definitions whereas object is defined inline?

bravado/mapping/response.py
analogue: Well, this class is meant to be subclassed and complain in a helpful way when the subclass' implementation is not fully compliant.

bravado/mapping/response.py
analogue: ack

bravado/mapping/unmarshal.py
analogue: generalized comment

bravado/mapping/unmarshal.py
analogue: Pretty much.  Models have to be in `#/definitions` and can be of any jsonschema type (not necessarily object). Not all objects are models, only the objects with the 'x-model' key which we insert manually as a pre-processing step in `model.tag_models(...)`

bravado/mapping/unmarshal.py
analogue: Maybe `x-model` was not the best name to use.  The use of this marker is entirely internal to bravado so it is not expected to be in the actual spec.  The purpose of the marker is to be able to differentiate between true `object` types which should map to a dict, and model types which map to a dynamically generated model type.  Since jsonref inlines `$ref`s to `#/definitions`, this is the only way to tell them apart.

PR #112, opened 2015-03-05
Add support for known 'format's

analogue:
Pretty self explanatory


bravado/mapping/formatter.py
```diff
@@ -1,66 +1,83 @@
...
+        lambda i: i if isinstance(i, int) else int(i),
+        lambda i: i if isinstance(i, int) else int(i),
+        'Converts integer:int32 <==> python int'),
+    'float': (
+        lambda f: f if isinstance(f, float) else float(f),
+        lambda f: f if isinstance(f, float) else float(f),
+        'Converts number:float <==> python float'),
+    'double': (
+        lambda d: d if isinstance(d, float) else float(d),
+        lambda d: d if isinstance(d, float) else float(d),
```
dnephin: I wonder if we should be using https://docs.python.org/2/library/decimal.html for double (or even float) so we don't lose precision.

dnephin:
lgtm, just one consideration about float vs decimal

bravado/mapping/formatter.py
analogue: I get how Decimal can cause the least surprise for a number of reasons but I think sticking float should suffice for now.  When/If issues arise we can revisit or maybe even add additional support for a 'decimal' format. Not sure how far that would go beyond the world of python but still a consideration.  I'll add your comment to the code so there is context for future readers.

dnephin: Switching it later is difficult because it's a non-backwards compatible change. We'd have to make a whole major version bump to get that in.

Since these are really just conventions, we could do float -> float, double -> Decimal ?

dnephin: Either way it doesn't block this merge, but maybe something we should decide before an official release.

PR #115, opened 2015-03-06
Remove dead code in bravado.swagger_type

analogue:
Highlights:
- `bravado/swagger_type.py` was mostly dead code and is now gone
- The still used funcs such as `is_dict_like()` and  `is_list_like()` were moved to `bravado/mapping/schema.py`
- `bravado.swagger_type.get_swagger_type()` was moved to `bravado/mapping/docstring.py` since it was only used for docstring purposes.  Also renamed it to `formatted_type()`
- move `SwaggerException` to `bravado.mapping.exception`


bravado/mapping/exception.py
```diff
@@ -0,0 +1,9 @@
+class SwaggerError(Exception):
+    """Raised when an error is encountered."""
+
+    def __init__(self, msg, cause=None):
+        """
+        :param msg: String message for the error.
+        :param cause: Optional exception that caused this one.
+        """
+        super(Exception, self).__init__(msg, cause)
```
dnephin: This should either go into `bravado/exception.py` or be renamed to be explicitly for mapping exceptions.

I think we already have this exact class in `bravado/exception.py`

bravado/mapping/schema.py
```diff
@@ -10,6 +11,14 @@
     'null': (),
 }
 
+SWAGGER_PRIMITIVES = (
+    'integer',
+    'number',
+    'string',
+    'boolean',
+    'null'     # TODO: Do we need this?
```
dnephin: according to the spec, this is correct

https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#data-types

references

http://json-schema.org/latest/json-schema-core.html#anchor8

Although I guess swagger special cases "array" and "object", and "file" is missing, but I guess that's a known deficiency right now? Do we have an issue open for "add file support for bravado" ?

dnephin:
Just one question about the exception class move

bravado/mapping/exception.py
analogue: I'm trying to reduce the coupling between `bravado` and `bravado.mapping` in hopes of making `bravado.mapping` easier to use in pyramid_swagger.  Currently all references to `SwaggerError` are in `bravado.mapping`.
```
spatel@vm-yelp:~/git/swagger-py-analogue (cleanup) $ git grep SwaggerError
bravado/mapping/docstring.py:from bravado.mapping.exception import SwaggerError
bravado/mapping/docstring.py:        raise SwaggerError("No proper type could be found from %s" % spec)
bravado/mapping/exception.py:class SwaggerError(Exception):
bravado/mapping/marshal.py:from bravado.mapping.exception import SwaggerError
bravado/mapping/marshal.py:        raise SwaggerError('Unknown type {0} for value {1}'.format(
bravado/mapping/operation.py:from bravado.mapping.exception import SwaggerError
bravado/mapping/operation.py:    :raises: SwaggerError when the status_code could not be mapped to a response
bravado/mapping/operation.py:        raise SwaggerError(
bravado/mapping/param.py:from bravado.mapping.exception import SwaggerError
bravado/mapping/param.py:        raise SwaggerError(
bravado/mapping/schema.py:from bravado.mapping.exception import SwaggerError
bravado/mapping/schema.py:                    raise SwaggerError(
bravado/mapping/unmarshal.py:from bravado.mapping.exception import SwaggerError
bravado/mapping/unmarshal.py:        raise SwaggerError(
tests/mapping/docstring/formatted_type_test.py:from bravado.mapping.exception import SwaggerError
tests/mapping/docstring/formatted_type_test.py:    with pytest.raises(SwaggerError) as excinfo:
tests/mapping/marshal/validate_primitive_test.py:# from bravado.exception import SwaggerError
tests/mapping/marshal/validate_primitive_test.py:#     with pytest.raises(SwaggerError) as excinfo:
tests/mapping/operation/get_response_spec_test.py:from bravado.mapping.exception import SwaggerError
tests/mapping/operation/get_response_spec_test.py:    with pytest.raises(SwaggerError) as excinfo:
tests/resource_response_test.py:from bravado.mapping.exception import SwaggerError
tests/resource_response_test.py:        self.assertRaises(SwaggerError, SwaggerClient.from_url,
```

I'd like to revisit specific exceptions at the `bravado` scope (connection but not mapping realted) eventually.

dnephin: Could we make this exception more explicitly named?

bravado/mapping/schema.py
analogue: - Removed TODO for the null.
- Added TODOs for support `file` type on marshal/unmarshal

bravado/mapping/exception.py
analogue: Fixed

dnephin:
lgtm once build is green

analogue:
Travis flaked out on this. Builds green locally.

PR #116, opened 2015-03-07
Validate arrays

analogue:
Implement Swagger Spec array validations when marshaling/unmarshaling array types.


dnephin:
lgtm

(with the reservation that I think doing validation in the client is unnecessary, and we should only worry about marshalling. As long as we're able to marshal from a type into the expected type, that is good enough for a client.)

PR #117, opened 2015-03-09
[bravado] Replace inbuilt async client with fido

prat0318:
Side note: As `fido` exposes `result()` on future and swagger-py also exposes `result()`, changes to `http_client` are made to expose `result()` which internally calls `wait()`.

It waits on Yelp/fido#2 to get merged in (some tests may be breaking right now).

PS: Not sure why `flake8` are breaking in `tests/mapping/schema/to_array_schema_test.py` as it is untouched in this commit.


bravado/http_client.py
```diff
@@ -197,6 +197,9 @@ def __init__(self, session, request):
         self.session = session
         self.request = request
 
+    def result(self, *args, **kwargs):
```
prat0318: This was added to unify interface of sync and async clients (as `fido` doesn't expose `wait` method).

bravado/response.py
```diff
@@ -69,9 +69,10 @@ def result(self, **kwargs):
 
         if self.cancelled():
             raise CancelledError()
-        response = self._request.wait(timeout=timeout)
+        response = self._request.result(timeout=timeout)
         try:
-            response.raise_for_status()
+            if hasattr(response, 'raise_for_status'):
```
prat0318: I am not sure if this is the correct way. `requests` silently stores errors when making the api call whereas fido throws up the exception during the api call. The `raise_for_status` is specific to `requests` to make it throw the exception if any.

bravado/async_http_client.py
```diff
@@ -44,116 +33,26 @@ def start_request(self, request_params):
...
-            body which is provided using deliverBody
-            """
-            response.deliverBody(_HTTPBodyFetcher(request_params,
-                                                  response, finished_resp))
-        deferred.addCallback(response_callback)
-
-        def response_errback(reason):
-            """Error callback method like server not reachable or conn. refused
+        if isinstance(url, unicode):
+            url = url.encode('utf-8')
```
dnephin: We should handle this in fido instead of here. Everything that calls fido is going to have to repeat this, which is unnecessary.

bravado/async_http_client.py
```diff
@@ -44,116 +33,26 @@ def start_request(self, request_params):
...
-                self.request, self.response, self.buffer.getvalue()))
-        else:
-            self.finished.errback(reason)
+        return fido.fetch(url, **request_params)
 
 
 def stringify_body(request_params):
     """Wraps the data using twisted FileBodyProducer
     """
+    data = None
```
dnephin: Instead of adding `data = None` can we change each of these to just return immediately?

```python

if ...:
    return create_multipart_content(...)

if headers.get(...):
    return ..

return param_stringify_body(...)

```

bravado/http_client.py
dnephin: The interface definition is not really clear in the current version (it's more of a base class). I think the this interface needs a lot of work, it doesn't have to happen right now (but it should happen before the bravado release).

A major problem with the current implementation is that it's mutable. There is no reason for it to be.

I think it should be something like this:

```python

RequestParams = namedtuple('RequestParams', ...)

class HttpClient(object):

    # this would replace start_request()
    def request(self, request_params):
        """Returns an HttpFuture"""

class HttpFuture(object):

    def result(self, timeout=DEFAULT_TIMEOUT):
        """Returns a swagger response, or raises an exception if there was an error making the request"""

```

bravado/response.py
dnephin: This should really be part of a specific Response implementation for each HttpClient.

Using the interfaces I describe above it would be easy to do this:

```python

class HttpFuture(object):

    def __init__(self, client, response_handler):
       ...

```

A `ResponseHandler` (currently called `post_receive`) can deal with the differences between each client, and since the client returns the Future, we don't need any branching. It know the correct class to return to handle the response.

bravado/mapping/operation.py
```diff
@@ -154,8 +156,9 @@ def handle_response(response, op):
     """
     if isinstance(response, requests.models.Response):
         wrapped_response = RequestsLibResponseAdapter(response)
+    elif isinstance(response, fido.fido.Response):
+        wrapped_response = FidoLibResponseAdapter(response)
```
dnephin: With my proposed interfaces we wouldn't need this branching.  The client would return the correct adaptor directly.

Branching is unnecessary, and type checks are not very pythonic.  We can avoid path.

bravado/swagger_model.py
```diff
@@ -40,6 +40,9 @@ def wait(self, timeout=None):
         with contextlib.closing(urllib.urlopen(self.get_path())) as fp:
             return self.FileResponse(json.load(fp))
 
+    def result(self, *args, **kwargs):
+        return self.wait(*args, **kwargs)
```
dnephin: Why?

bravado/async_http_client.py
prat0318: Added Yelp/fido#3 regarding it.

prat0318: On second thought, After execution of `url = '%s?%s' % ...` above, instance of `url` will always be `str`, isnt it? So lines 46-47 can anyways be safely removed.

bravado/async_http_client.py
prat0318: done

bravado/async_http_client.py
dnephin: Nope, in python2.x if the variables used in the substitution are unicode, the final result will be unicode as well:

```python

>>> '%s' % (u'asdf')
u'asdf'
```

prat0318: ah, gotcha.

bravado/swagger_model.py
prat0318: This is because `FileEventual` [could also be a result](https://github.com/Yelp/swagger-py/blob/swagger2.0/bravado/swagger_model.py#L54-55) of `start_request`, so had to add `result()` to its interface as well.

bravado/http_client.py
prat0318: I will create a separate ticket for the refactoring.

bravado/async_http_client.py
prat0318: Will use the bumped version of fido PR once Yelp/fido#4 gets merged.

bravado/async_http_client.py
```diff
@@ -44,111 +33,16 @@ def start_request(self, request_params):
```
analogue: Github won't let me place this comment where I want (line 34 pre-commit, line 23 post-commit) but I think we need to stop beating around the bush and creating additional "naming" layers where they're not necessary.  So:

s/AsynchronousHttpClient/FidoHttpClient/g
s/SynchronousHttpClient/RequestsHttpClient/g

Ok to defer to #120

bravado/http_client.py
analogue: +1.  I'd like to divorce any specific HttpClient implementation from `bravado.mapping` in preparation for splitting out to a separate python library (more details to come).   Fido and Requests specifics should not be in `bravado.mapping`.

bravado/response.py
```diff
@@ -93,3 +94,20 @@ def status_code(self):
 
     def json(self, **kwargs):
         return self._delegate.json(**kwargs)
+
+
+class FidoLibResponseAdapter(ResponseLike):
```
analogue: s/FidoLib/Fido

The only request I used `RequestsLib` is because `Requests` is kind of very generic all by itself and not obvious that it referes to the "Request Library".

analogue:
Also, move HttpClient + friends into `bravado.mapping` or defer until #120.  

lgtm

bravado/async_http_client.py
dnephin: I like that, it'll make it easier to add other clients without name conflicts

dnephin:
>  move HttpClient + friends into `bravado.mapping`

I thought `mapping` was the package for things that are related to mapping swagger spec to a request , not related to the client?

analogue:
@dnephin Ideally, I would like that to be the case too. However, I think it is going to be hard to decouple the two (mapping from the http client) w/o the interface for the http client residing in `bravado.mappping`.

dnephin:
Ah, I think I misunderstood originally. That seems fine (interface in mapping, implementation in `bravado.http` or something like that).

bravado/async_http_client.py
prat0318: Will take it up in #120

bravado/response.py
prat0318: yeah, basically i wanted to keep both the namings consistent. I am changing `RequestsLibResponse..` to `RequestsResponse..` as well.

bravado/http_client.py
prat0318: @dnephin can you explain a bit on which part of the implementation is mutable here. I'm kinda lost about it.

dnephin: Hmm, I thought it used to be, but I don't see it any more, maybe it's been fixed already?

I guess the only issue now is the need for the branch logic, instead of passing along a Future that already has the correct adapter from the `request()` call.

prat0318: So, in `operation.py` [L156](https://github.com/Yelp/swagger-py/blob/swagger2.0/bravado/mapping/operation.py#L156) instead of returning the `__call()__` as:
```python
       return HTTPFuture(self.swagger_spec.http_client, request, response_future)
```
should the signature change to : 
```python
    return self.swagger_spec.http_client.request(request, response_future)
```

Which means, `http_client`'s `request()` should now return back a `HttpFuture` object?

dnephin: Something like that, we should be able to delete `handle_response` from the module, so it would probably be more like:

```python
return self.swagger_spec.http_client.request(request, self)
```

which lets us remove the `response_future` as well (it's just a closure so we have access to the operation anyway).

PR #121, opened 2015-03-13
Add support for additionalProperties to object + Model marshaling/unmarshaling

analogue:
This brings `additionalProperties` support to both objects and models - See https://github.com/swagger-api/swagger-spec/issues/285 

Pay close attn to `None` handling. The behavior can be summarized as:
- json nulls can't be sent over the wire unless the jsonschema explicitly specificies the type as `null`
- Python primitives with a value of `None` will not be present in the on the wire json representation
- bravado marshaling behavior will automatically take care of re-introducing properties that were absent from the on the wire format because of a `None` value iff that property was explicitly defined (was not an `additionalProperty`).

^ is sure to be confusing - we can discuss further if necessary.

Also, keep in mind `additionalProperties` can be either a boolean or a schema.


dnephin:
lgtm

PR #123, opened 2015-03-17
Add 'file' support to parameters

analogue:
Useful documentation for the Requests lib on file support - http://docs.python-requests.org/en/latest/user/advanced/#post-multiple-multipart-encoded-files

### DIFF
--- a/bravado/mapping/operation.py
+++ b/bravado/mapping/operation.py
@@ -35,6 +35,18 @@ class Operation(object):
         # (key, value) = (param name, Param)
         self.params = {}
 
+    @property
+    def consumes(self):
+        """
+        :return: List of supported mime types consumed by this operation. e.g.
+            ["application/x-www-form-urlencoded"]
+        :rtype: list of strings, never None
+        """
+        result = self.op_spec.get('consumes')
+        if result is None:
+            result = self.swagger_spec.spec_dict.get('consumes', [])
+        return result
+
     @classmethod
     def from_spec(cls, swagger_spec, path_name, http_method, op_spec):
         """
@@ -63,7 +75,7 @@ class Operation(object):
         param_specs = op_param_specs + path_param_specs
 
         for param_spec in param_specs:
-            param = Param(self.swagger_spec, param_spec)
+            param = Param(self.swagger_spec, self, param_spec)
             self.params[param.name] = param
 
     @property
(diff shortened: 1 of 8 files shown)


bravado/mapping/operation.py
```diff
@@ -63,7 +75,7 @@ def build_params(self):
         param_specs = op_param_specs + path_param_specs
 
         for param_spec in param_specs:
-            param = Param(self.swagger_spec, param_spec)
+            param = Param(self.swagger_spec, self, param_spec)
```
prat0318: is it totally necessary to send the op in the request? how about just the `consumes` field?

analogue: `operation_id` is also used. I predict `produces` will also be used at some point too.

bravado/mapping/param.py
```diff
@@ -108,3 +112,44 @@ def marshal_param(param, value, request):
...
+    if request.get('file') is None:
+        # support multiple files by default by setting to an empty array
+        request['file'] = []
+
+        # The http client should take care of setting the content-type header
+        # to 'multipart/form-data'. Just verify that the swagger spec is
+        # conformant
+        expected_mime_type = 'multipart/form-data'
+        if expected_mime_type not in param.op.consumes:
+            raise SwaggerMappingError((
```
prat0318: This should ideally be the job of rich validations of `swagger_spec_validator`.

prat0318: oh my bad. `consumes` is a part of the `request` and cant be checked with ssv.

dnephin: I think you're right, this should be part of the static validation of the schema, performed by ssv.

bravado/mapping/param.py
```diff
@@ -108,3 +112,44 @@ def marshal_param(param, value, request):
...
+def add_formdata(param, value, request):
+    """Add a parameter as formdata to the given request.
+
+    :type param: :class;`bravado.mapping.param.Param`
+    :param value: parameter value
+    :type request: dict
+    """
+    if request.get('data') is None:
+        request['data'] = {}
+    request['data'][param.name] = value
```
dnephin: Minor: you can do this as:

```python
request.setdetault('data', {})[param.name] = value
```

bravado/mapping/param.py
analogue: Add TODO and ticketed ssv

bravado/mapping/param.py
analogue: Fixed

PR #124, opened 2015-03-18
Remove caching from the client

analogue:
We've talked about doing this previously. Caching is hard and something the client needs to be able to reason about to be implemented correctly.


dnephin:
lgtm

PR #127, opened 2015-03-23
Refactore HttpClient and HttpFuture

### LINKED ISSUE #120: Refactor `HttpClient` and `HttpFuture`
Refactor `bravado/mapping/operation.py` such that `response_future` is all done by specific `HttpClient` and we can get rid of type checking. Refer #117 comments for details.

prat0318:
- Fixes #120 (Most of the implementation taken from comments of #117)
- Interface `HttpClient` and `HttpFuture` are place in `bravado.mapping`
- Implementations are kept in `bravado/`
- Removed `SynchronousEventual` and `handle_respone` tests

Pending: 
- I have yet to write more unit tests to it, but did not want the PR to get more bulkier.
- More functional tests.


bravado/fido_client.py
```diff
@@ -42,7 +39,9 @@ def start_request(self, request_params):
             'headers': request_params.get('headers', {}),
         }
 
-        return fido.fetch(url, **request_params)
+        fido_future = FidoFuture(op, url, request_params)
+        fido_future.fetch()
+        return fido_future
```
analogue: docstring and return type don't match. Also add `op` to docstring

bravado/fido_future.py
```diff
@@ -0,0 +1,69 @@
...
+    """
+    def __init__(self, requests_lib_response):
+        self._delegate = requests_lib_response
+
+    @property
+    def status_code(self):
+        return self._delegate.code
+
+    def json(self, **_):
+        return self._delegate.json()
```
analogue: Pass the `**kwargs`

bravado/mapping/operation.py
```diff
@@ -1,14 +1,14 @@
+#!/usr/bin/env python
```
analogue: ??

bravado/mapping/http_future.py
```diff
@@ -0,0 +1,19 @@
...
+
+#
+# Copyright (c) 2013, Digium, Inc.
+# Copyright (c) 2014, Yelp, Inc.
+#
+
+DEFAULT_TIMEOUT_S = 5.0
+
+
+class HttpFuture(object):
```
analogue: This can just be in http_client.py unless you've got a good reason to make it a separate module.

tests/async_http_client_test.py
```diff
@@ -15,19 +15,19 @@
...
 from ordereddict import OrderedDict
 
-import bravado.async_http_client
+import bravado.fido_client
 import bravado.exception
-import bravado.http_client
+from bravado.mapping.http_client import APP_FORM
 
 
 class AsyncHttpClientTest(unittest.TestCase):
```
analogue: Rename

tests/mapping/operation/unmarshal_response_test.py
```diff
@@ -9,8 +9,10 @@ def test_no_content(empty_swagger_spec):
         'description': "I don't have a 'schema' key so I return nothing",
     }
     response = Mock(spec=ResponseLike, status_code=200)
-    status_code, value = unmarshal_response(
-        empty_swagger_spec, response_spec, response)
+    with patch('bravado.mapping.operation.get_response_spec') as m:
+        m.return_value = response_spec
+        op = Mock(swagger_spec=empty_swagger_spec)
```
analogue: `spec=`

tests/mapping/operation/unmarshal_response_test.py
```diff
@@ -22,12 +24,15 @@ def test_json_content(empty_swagger_spec):
             'type': 'string',
         }
     }
+    op = Mock(swagger_spec=empty_swagger_spec)
```
analogue: spec=

tests/mapping/operation/unmarshal_response_test.py
```diff
@@ -22,12 +24,15 @@ def test_json_content(empty_swagger_spec):
...
     response = Mock(
         spec=ResponseLike,
         status_code=200,
         json=Mock(return_value='Monday'))
 
-    status_code, value = unmarshal_response(
-        empty_swagger_spec, response_spec, response)
+    with patch('bravado.mapping.operation.get_response_spec') as m:
+        m.return_value = response_spec
+        op = Mock(swagger_spec=empty_swagger_spec)
```
analogue: spec=

analogue:
lgtm except for the minor stuff

bravado/fido_client.py
```diff
@@ -2,29 +2,26 @@
...
 
 
-class AsynchronousHttpClient(http_client.HttpClient):
-    """Asynchronous HTTP client implementation.
+class FidoClient(HttpClient):
+    """Fido HTTP client implementation.
     """
 
-    def start_request(self, request_params):
+    def request(self, request_params, op=None):
```
dnephin: Since operation is required, I would make it a regular argument, not a kwarg.

prat0318: This is because `op` is not required for api-docs fetch. For that instance, `client.request(request_params)` has to be called without `op`.

bravado/fido_future.py
```diff
@@ -0,0 +1,69 @@
...
+    """A future which inputs HTTP params"""
+
+    def __init__(self, op, url, request):
+        """Kicks API call for Fido client
+
+        :param request: dict containing API request parameters
+        """
+        self.op = op
+        self.url = url
+        self.request = request
```
dnephin: I don't think we need inheritance here. We can have a single `HttpFuture` that takes a `ResponseAdapter`

See example below

bravado/fido_future.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def fetch(self):
+        self.future = fido.fetch(self.url, **self.request)
```
dnephin: This is mutable! It doesn't need to be.

Why not pass in the future to the constructor, so in the `FidoClient` you'd have:

```python
return HttpFuture(op, FidoResponseAdapter , fido.fetch(url, request_params))
```

The request is only used after this for debug logging. I'm sure we can do that logging in fido instead, it doesn't need to be here.

bravado/mapping/http_client.py
```diff
@@ -0,0 +1,106 @@
...
+        """Configures client to use api_key authentication.
+
+        The api_key is added to every query parameter sent.
+
+        :param host: Hostname to limit authentication to.
+        :param api_key: Value for api_key.
+        :param param_name: Parameter name to use in query string.
+        """
+        raise NotImplementedError(
+            u"%s: Method not implemented", self.__class__.__name__)
```
dnephin: The above two methods (`set_basic_auth` and `set_api_key`) are not actually part of the interface. They are never implemented by the fido client (and I don't think they should be).

Can you remove them from this interface class?

bravado/requests_client.py
```diff
@@ -0,0 +1,75 @@
...
+            self.session,
+            self.authenticated_request(request_params))
+
+    def set_basic_auth(self, host, username, password):
+        self.authenticator = BasicAuthenticator(
+            host=host, username=username, password=password)
+
+    def set_api_key(self, host, api_key, param_name=u'api_key'):
+        self.authenticator = ApiKeyAuthenticator(
+            host=host, api_key=api_key, param_name=param_name)
```
dnephin: It would  be great if we could do this from the constructor we this class could be immutable as well, but that can wait for some other time.

bravado/requests_future.py
```diff
@@ -0,0 +1,85 @@
...
+            timeout=timeout)
+
+        self.check_for_exceptions(response)
+
+        # op is not present during api-docs fetch, return raw response
+        if not self.op:
+            return response
+
+        return unmarshal_response(RequestsResponseAdapter(response),
+                                  self.op)
```
dnephin: If this class goes away (so that we can just use a single generic `HttpFuture` that accepts the response adapter), I think you could implement this as:

```python
def result(self, timeout...):
    ...
    response = self.future.result(timeout=timeout)
    return unmarshal_response(self.response_adapter(response), self.op)
```

And in `RequestsClient.request` you just need a `FutureAdapter` on the request

```python
return HttpFuture(
    op,
    RequestsResponseAdapter,
    FutureAdapter(self.session,  self.authenticated_request(request_params))
```

Where `FutureAdapter` would be something like this

```python

class FutureAdapter(object):

def __init__(self, session, request):
    ...

def result(self, timeout=timeout):
    return self.session.send(self.session.prepare_request(request), timeout=timeout)
```

bravado/fido_client.py
dnephin: Ah, ok, fair enough, kwargs works

dnephin:
This is looks good, I think with a few more small changes we can remove the mutability and inheritance as well

bravado/requests_future.py
```diff
@@ -0,0 +1,85 @@
...
+        log.debug(u"%s %s(%r)", request.method, request.url, request.params)
+        response = self.session.send(
+            self.session.prepare_request(request),
+            timeout=timeout)
+
+        self.check_for_exceptions(response)
+
+        # op is not present during api-docs fetch, return raw response
+        if not self.op:
+            return response
```
dnephin: I think having very different responses based on the presence of an operation is kind of unexpected. 

I also just noticed that self.op is only being used the unmarshal the response.

It seems like this interface would be much more concise if `unmarshal_response(response, op)` were actually called by whatever calls this.

That would allow us to remove self.op as a parameter entirely.

dnephin: Ya, I think we can accomplish this by restoring `handle_response(response, op)`, but without any of the branch logic. Since the adapters normalize the response to a single representation.

bravado/fido_future.py
prat0318: but `Fido`'s response doesn't take any params to json() in [fido.py](https://github.com/Yelp/fido/blob/master/fido/fido.py#L41)

bravado/mapping/operation.py
prat0318: yeah, saw it in swagger-py files. will remove them.

bravado/fido_future.py
prat0318: Ack. Added fido logging as an issue in Yelp/fido#5

bravado/mapping/http_future.py
prat0318: This was to keep the interfaces easy to access: `http_client` -> `requests_client` & `fido_client` and `http_future` -> `requests_future` & `fido_future`.

tests/async_http_client_test.py
prat0318: Ack.

bravado/mapping/http_client.py
prat0318: Ack.

bravado/requests_client.py
prat0318: Added an issue #130

bravado/requests_future.py
prat0318: As discussed, ignoring this to instead callback to the clients.

bravado/requests_future.py
prat0318: Ack.

bravado/requests_future.py
dnephin: I think even with the callback you'd need this adapter, no?

bravado/fido_client.py
prat0318: ack.

bravado/mapping/http_future.py
prat0318: HttpFuture is now a class instead of an interface. Keeping it in `http_future.py`.

prat0318:
- Refactored to re-introduce callbacks.
- Made HttpFuture as a class instead of an interface.

@analogue I saw that in the integration tests, `.body` and `.text` are expected from the `result` but the `ResponseLike` only exposes `.json` and `.status_code`. I am not sure why it was not throwing up earlier, but now the tests throw `No attribute present body` etc.

Also, functional tests fail with `jsonschema` validation failure of parameter count being 0.

tests/mapping/operation/unmarshal_response_test.py
prat0318: It is a mock of `op` hence added the attribute `swagger_spec`.

bravado/mapping/http_future.py
dnephin: I think this needs to be moved out of `bravado/mapping` though into just `bravado`. This would never be used by the server.

bravado/mapping/http_future.py
```diff
@@ -0,0 +1,40 @@
...
+    def __init__(self, request_adapter, response_adapter, callback):
+        """Kicks API call for Fido client
+
+        :param request_adapter: Adapter which exposes result() method
+        of the http client
+        :param response_adapter: Adapter which exposes json(), status_code()
+        :type response_adapter: :class: `bravado.mapping.response.ResponseLike`
+        :param callback: Function to be called on the response
+        """
+        self.request_adapter = request_adapter
```
dnephin: s/request_adapter/future/ , since the interface is `concurrent.futures.Future`

bravado/mapping/operation.py
```diff
@@ -153,45 +149,35 @@ def __call__(self, **kwargs):
         log.debug(u"%s(%s)" % (self.operation_id, kwargs))
         request = self.construct_request(**kwargs)
 
-        def response_future(response, **kwargs):
-            return handle_response(response, self, **kwargs)
+        def response_callback(response_adapter):
+            return handle_response(response_adapter, self)
```
dnephin: I guess since `handle_response()` just calls `unmarshal_response()`, we can just call unmarshal_response here directly?

bravado/requests_client.py
```diff
@@ -188,28 +129,70 @@ def apply_authentication(self, request):
...
+        """Blocking call to wait for API response
 
-        :param timeout: timeout for the request, in seconds
+        :param timeout: timeout in seconds to wait for response
+        :type timeout: integer
+        :return: raw response from the server
+        :rtype: dict
         """
         request = self.request
         log.debug(u"%s %s(%r)", request.method, request.url, request.params)
```
dnephin: I think this `log.debug()` is unnecessary, since the requests module already has very same logging.

dnephin:
This looks great! Thanks for making all the changes.  Just some minor rename/docstring stuff left

bravado/mapping/http_future.py
dnephin: Although we can do that later when we go to extract this package

bravado/mapping/operation.py
prat0318: yeah, true.

bravado/mapping/http_future.py
prat0318: ack

bravado/requests_client.py
prat0318: will try that out on repl. will remove if it is unnecessary.

prat0318: Turns out that only `hostname` is not present from the `requests` debug, else it looks more or less the same:
```
DEBUG:post http://petstore.swagger.io/v2/pet({})
# below is requests debug log
DEBUG:"POST /v2/pet HTTP/1.1" 200 None
```

prat0318:
@analogue I am adding `text` property to requests and fido adapters which contain the raw response from the server. This is to make the new integration tests pass. It is kinda other way round (to make the tests pass), but i think it is ok to have a raw response option available to the client.

bravado/requests_client.py
dnephin: Yup, I think even the hostname is usually part of an earlier log line where it establishes the connection for the session, so it's pretty easy to reconcile them.

bravado/fido_future.py
analogue: Ticket `fido.py` and add a TODO here.  We really shouldn't impose such a constraint.

bravado/mapping/http_future.py
analogue: Agreed just as long as we don't have any references to it in `bravado.mapping`.

tests/mapping/operation/unmarshal_response_test.py
analogue: See the purpose of `class Mock(spec=None, ...` @ http://www.voidspace.org.uk/python/mock/mock.html

Two most important reasons this is a good practice:
- Let's future readers know exactly what type of object is being mocked
- Makes sure the unit test doesn't deviate from the type's interface

bravado/exception.py
```diff
@@ -2,7 +2,7 @@
 # -*- coding: utf-8 -*-
```
analogue: looks like there are some remaining

bravado/fido_client.py
```diff
@@ -1,37 +1,59 @@
...
-class AsynchronousHttpClient(http_client.HttpClient):
-    """Asynchronous HTTP client implementation.
+class FidoResponseAdapter(ResponseLike):
+    """Wraps a fido.fido.Response object to provider a uniform interface
+    to the response innards.
+
+    :type requests_lib_response: :class:`fido.fido.Response`
+    """
+    def __init__(self, requests_lib_response):
+        self._delegate = requests_lib_response
```
analogue: fido_response

bravado/fido_client.py
```diff
@@ -1,37 +1,59 @@
...
+    @property
+    def text(self):
+        return self._delegate.body
+
+    def json(self, **_):
+        return self._delegate.json()
+
+
+class FidoClient(HttpClient):
+    """Fido HTTP client implementation.
```
analogue: Insert "async" somewhere in the docs so it is obvious how it differs from `RequestsClient`

bravado/fido_client.py
```diff
@@ -1,37 +1,59 @@
...
         Sets up crochet and triggers the API request in background
 
         :param request_params: request parameters for API call
         :type request_params: dict
+        :param response_callback: Function to be called after
+        receiving the response
+        :type response_callback: method
 
-        :return: crochet EventualResult
+        :return: :class: `bravado.mapping.http_future.HttpFuture`
```
analogue: :rtype:

bravado/mapping/http_client.py
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env python
```
analogue: another one

tests/mapping/operation/unmarshal_response_test.py
```diff
@@ -54,6 +63,9 @@ def test_performs_validation(empty_swagger_spec, response_spec):
...
 
-    with patch('bravado.mapping.operation.validate_schema_object') as m:
-        unmarshal_response(empty_swagger_spec, response_spec, response)
-    assert m.call_count == 1
+    with patch('bravado.mapping.operation.validate_schema_object') as m1:
+        with patch('bravado.mapping.operation.get_response_spec') as m:
+            m.return_value = response_spec
+            op = Mock(swagger_spec=empty_swagger_spec)
+            unmarshal_response(response, op)
+            assert m1.call_count == 1
```
analogue: meaningful name?

analogue:
`ResponseLike` needs to updated to support the `text` attr

bravado/fido_future.py
prat0318: Ack.

bravado/exception.py
prat0318: Removed the line: `#!/usr/bin/env python`. 
Shouldn't we define the python encoding as `utf-8` instead of default ascii as the best practice?

bravado/fido_client.py
prat0318: ah right.

bravado/fido_client.py
prat0318: yup.

bravado/fido_client.py
prat0318: ack.

bravado/mapping/http_client.py
prat0318: ack.

tests/mapping/operation/unmarshal_response_test.py
prat0318: ack.

analogue:
lgtm! good job

PR #128, opened 2015-03-23
When marshaling a request on the client side or a response on the server...

analogue:
... side, allow dicts to be passed in for models

### DIFF
--- a/bravado/mapping/marshal.py
+++ b/bravado/mapping/marshal.py
@@ -31,9 +31,14 @@ def marshal_schema_object(swagger_spec, schema_object_spec, value):
         return marshal_array(swagger_spec, schema_object_spec, value)
 
     if is_model(schema_object_spec):
-        # It is important that the 'model' check comes before 'object' check.
-        # Model specs also have type 'object' but also have the additional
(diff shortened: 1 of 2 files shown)


bravado/mapping/marshal.py
```diff
@@ -31,9 +31,14 @@ def marshal_schema_object(swagger_spec, schema_object_spec, value):
...
-        # MODEL_MARKER key for identification.
+
+        # Allow models to be passed in as dicts for flexibility.
+        if is_dict_like(value):
+            return marshal_object(swagger_spec, schema_object_spec, value)
+
+        # It is important that the 'model' check comes before 'object' check
+        # below. Model specs are of type 'object' but also have a MODEL_MARKER
+        # key for identification.
         return marshal_model(swagger_spec, schema_object_spec, value)
```
dnephin: Minor: Took me a while to see the difference between these branches.  Could be written as:

```python

marshal_func = marshal_object if is_dict_like(value) else marshal_model
marshal_func(swagger_spec, schema_object_spec, value)
```

which makes it more obvious

dnephin:
One minor suggestion, but lgtm either way

PR #129, opened 2015-03-23
Make response validation configurable

analogue:


bravado/mapping/spec.py
```diff
@@ -5,27 +6,39 @@
...
+        :param origin_url: URL from which the spec was retrieved.
+        :param http_client: :class:`bravado.http_client.HTTPClient`
+        :param config: Configuration dict. See CONFIG_DEFAULTS.
+        """
         self.spec_dict = spec_dict
         self.origin_url = origin_url or 'unknown'
         self.http_client = http_client
         self.api_url = None
+        self.config = copy.deepcopy(CONFIG_DEFAULTS)
+        self.config.update(config or {})
```
dnephin: `deepcopy` is one of the worse performance offenders. We should never need to deepcopy.

Why not do this instead:

```python
self.config = dict(CONFIG_DEFAULTS, **(config or {}))
```

which is pythons version of "merge dicts"

PR #131, opened 2015-03-24
Patch to allow POST with binary file in python 2.7

bkbarry:
When Resources/Operations are constructed from api-docs, all fields
(e.g. 'method') are unicode. In python 2.7 httplib attempts string
concatenation to construct POST bodies, which means that
UnicodeDecodeErrors occur if one includes binary data as a parameter.
Rather than change all api-doc fields to 'str', this branch casts only
fields used in httplib.

Might be prudent to file a followup issue to convert all test resource listings to unicode fields, since this is what the requests library returns.

### DIFF
--- a/swaggerpy/client.py
+++ b/swaggerpy/client.py
@@ -180,7 +180,7 @@ class Operation(object):
         request = {}
         # The requests library expects native strings. Without this, POST
         # with a binary file upload throws a UnicodeDecodeError.
-        request['method'] = str(self._json[u'method'])
+        request['method'] = self._json[u'method'].encode('utf8')
         request['url'] = self._uri
         request['params'] = {}
         request['headers'] = _request_options.get('headers', {}) or {}
--- a/swaggerpy/client.py
+++ b/swaggerpy/client.py
@@ -178,6 +178,8 @@ class Operation(object):
         _request_options = kwargs.pop('_request_options', {}) or {}
 
         request = {}
+        # The requests library expects native strings. Without this, POST
+        # with a binary file upload throws a UnicodeDecodeError.
         request['method'] = str(self._json[u'method'])
         request['url'] = self._uri
         request['params'] = {}
--- a/swaggerpy/client.py
+++ b/swaggerpy/client.py
@@ -178,7 +178,7 @@ class Operation(object):
         _request_options = kwargs.pop('_request_options', {}) or {}
 
         request = {}
-        request['method'] = self._json[u'method']
+        request['method'] = str(self._json[u'method'])
         request['url'] = self._uri
         request['params'] = {}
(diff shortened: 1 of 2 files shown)


tests/client_test.py
```diff
@@ -328,6 +343,28 @@ def setUp(self):
                                         ]
                                     }
                                 ]
+                            },
+                            {
+                                "path": "/pet/{petId}/vaccine",
+                                "operations": [
+                                    {
+                                        "method": u"POST",
```
johnschultz: Perhaps a comment about why `u"POST"` instead of just `"POST"`?

swaggerpy/client.py
```diff
@@ -178,7 +178,7 @@ def _construct_request(self, **kwargs):
         _request_options = kwargs.pop('_request_options', {}) or {}
 
         request = {}
-        request['method'] = self._json[u'method']
+        request['method'] = str(self._json[u'method'])
```
johnschultz: Perhaps a note as to why the cast?

johnschultz:
Aside from those two terribly minor nits, this looks good to me. I'll defer to someone who contributes more to this project hit the merge button. 

Though it is so big and green...

sjaensch:
Some context: this is a known issue for the requests library, which the synchronous http client uses: https://github.com/kennethreitz/requests/issues/1723

Also note that at the end of that issue it's explicitly stated that requests expects native strings (str in Python 2) everywhere.

I'd appreciate a quick merge since this is currently blocking our transition to Python 2.7.

tests/client_test.py
bkbarry: Will do. Really, all of these should be unicode since that's what Response.json() returns. Going to cut a followup branch once we have a little more bandwidth.

swaggerpy/client.py
dnephin: I believe the correct way to do this is actually `.encode('utf8')` ?

bkbarry: Hey @dnephin - for my own edification, what's the reasoning behind that? Just that str() uses the default encoding, which can vary between environments? If so, we should probably change the async client since it also casts unicode fields to strings with str().

dnephin:
@bkbarry your question is now unlinked from this PR because of a rebase I think? But the short answer is, I don't remember why `.encode('utf8')` is preferred, or if I'm wrong about that.

I would guess it has something to do with being explicit about the encoding? If you just do `str()` it just grabs whatever happens to be default.  I think buck would have a better answer, and maybe even tell me I'm wrong.

podarsmarty:
the .encode('utf8') cause problems once you go into python 3 since we now have type byte being returned instead of a str with the changes to unicode/str.

PR #135, opened 2015-03-29
Unifying exceptions between swaggerpy and requests

laucia:
fix for https://github.com/Yelp/swagger-py/issues/134


prat0318:
It doesn't seem to be a good way to limit the common HTTPClient to the `requests` one, but instead it should be the one from `swaggerpy`.

PR #136, opened 2015-03-31
Add Swagger20Validator - an extenstion of jsonschema's Draft4Validator

### LINKED ISSUE #126: [bravado] jsonschema.validate is failing on request validation
I tried this:

```python
swagger_client = client.get_client('http://petstore.swagger.io/v2/swagger.json')
swagger_client.pet.findPetsByStatus(status=['available']).result()
```

Which throws back:
```
---> 33     jsonschema.validate(value, spec)
     34
     35

/usr/lib/pymodules/python2.6/jsonschema/validators.pyc in validate(instance, schema, cls, *args, **kwargs)
    504     """
    505     if cls is None:
    506         cls = validator_for(schema)
--> 507     cls.check_schema(schema)
    508     cls(schema, *args, **kwargs).validate(instance)

/usr/lib/pymodules/python2.6/jsonschema/validators.pyc in check_schema(cls, schema)
     74         def check_schema(cls, schema):
     75             for error in cls(cls.META_SCHEMA).iter_errors(schema):
---> 76                 raise SchemaError.create_from(error)
     77
     78         def iter_errors(self, instance, _schema=None):

SchemaError: False is not of type u'array'

Failed validating u'type' in schema[u'properties'][u'required']:
    {u'items': {u'type': u'string'},
     u'minItems': 1,
     u'type': u'array',
     u'uniqueItems': True}

On instance[u'required']:
    False
```

It looks like `swagger_spec` can not be directly treated as `json_schema` to validate the request.

On removing `required` field from the spec, and retrying gives:
```
ValidationError: ['available'] is not one of [u'available', u'pending', u'sold']

Failed validating u'enum' in schema:
    {u'collectionFormat': u'multi',
     u'default': u'available',
     u'description': u'Status values that need to be considered for filter',
     u'enum': [u'available', u'pending', u'sold'],
     u'in': u'query',
     u'items': {u'type': u'string'},
     u'name': u'status',
     u'type': u'array'}

On instance:
    ['available']
```

In other words, `jsonschema.validate` will not work in these cases.

prat0318:
Fixes #126.

- Added the cases i found were breaking earlier. Will add more extensions later on to the validator.

### DIFF
--- a/bravado/mapping/swagger20_validator.py
+++ b/bravado/mapping/swagger20_validator.py
@@ -2,6 +2,14 @@ from jsonschema import validators, _validators
 from jsonschema.exceptions import ValidationError
 from jsonschema.validators import Draft4Validator
 
+from bravado.mapping.schema import is_param_spec
+
+"""Draft4Validator is not completely compatible with Swagger 2.0 schema
+objects like parameter, etc. Swagger20Validator is an extension of
+Draft4Validator which customizes/wraps some of the operations of the default
+validator.
+"""
+
 
 def ignore(_validator, *args):
     """A validator which performs no validation. Used to `ignore` some schema
@@ -13,10 +21,18 @@ def ignore(_validator, *args):
 def required_validator(validator, req, instance, schema):
     """Swagger 2.0 expects `required` to be a bool in the Parameter object,
     but a list of properties everywhere else.
+
+    :param validator: Validator class used to validate the object
+    :type validator: :class: `Swagger20Validator` or
+                             `jsonschema.validators.Draft4Validator`
+    :param req: value of `required` field
+    :type req: boolean or list or None
+    :param instance: object instance value
+    :param schema: swagger spec for the object
+    :type schema: dict
     """
-    # TODO: Find a better way to identify if it is a Parameter schema
-    if schema.get('in'):
-        if req is True and not instance:
+    if is_param_spec(schema):
+        if req is True and instance is None:
             return [ValidationError("%s is required" % schema['name'])]
     else:
         return _validators.required_draft4(validator, req, instance, schema)
@@ -25,13 +41,22 @@ def required_validator(validator, req, instance, schema):
 def enum_validator(validator, enums, instance, schema):
     """Swagger 2.0 allows enums to be validated against objects of type
     arrays, like query parameter (collectionFormat: multi)
+
+    :param validator: Validator class used to validate the object
+    :type validator: :class: `Swagger20Validator` or
+                             `jsonschema.validators.Draft4Validator`
+    :param enums: allowed enum values
+    :type enums: list
+    :param instance: enum instance value
+    :param schema: swagger spec for the object
+    :type schema: dict
     """
     if schema.get('type') == 'array':
         return (v for item in instance for v in _validators.enum(
             validator, enums, item, schema))
     return _validators.enum(validator, enums, instance, schema)
 
-Swagger20RequestValidator = validators.extend(
+Swagger20Validator = validators.extend(
     Draft4Validator,
     {
         'required': required_validator,
--- a/bravado/mapping/swagger20_validator.py
+++ b/bravado/mapping/swagger20_validator.py
@@ -1,0 +1,39 @@
+from jsonschema import validators, _validators
+from jsonschema.exceptions import ValidationError
+from jsonschema.validators import Draft4Validator
+
+
+def ignore(_validator, *args):
+    """A validator which performs no validation. Used to `ignore` some schema
+    fields during validation.
+    """
+    return
+
+
+def required_validator(validator, req, instance, schema):
+    """Swagger 2.0 expects `required` to be a bool in the Parameter object,
+    but a list of properties everywhere else.
+    """
+    # TODO: Find a better way to identify if it is a Parameter schema
+    if schema.get('in'):
+        if req is True and not instance:
+            return [ValidationError("%s is required" % schema['name'])]
+    else:
+        return _validators.required_draft4(validator, req, instance, schema)
+
+
+def enum_validator(validator, enums, instance, schema):
+    """Swagger 2.0 allows enums to be validated against objects of type
+    arrays, like query parameter (collectionFormat: multi)
+    """
+    if schema.get('type') == 'array':
+        return (v for item in instance for v in _validators.enum(
+            validator, enums, item, schema))
+    return _validators.enum(validator, enums, instance, schema)
+
+Swagger20RequestValidator = validators.extend(
+    Draft4Validator,
+    {
+        'required': required_validator,
+        'enum': enum_validator,
+    })
--- a/bravado/mapping/validate.py
+++ b/bravado/mapping/validate.py
@@ -5,7 +5,7 @@ customize the behavior.
 """
 from bravado.mapping.exception import SwaggerMappingError
 from bravado.mapping.schema import SWAGGER_PRIMITIVES
-from bravado.mapping.swagger20_validator import Swagger20RequestValidator
+from bravado.mapping.swagger20_validator import Swagger20Validator
 
 
 def validate_schema_object(spec, value):
@@ -33,7 +33,7 @@ def validate_primitive(spec, value):
     :param spec: spec for a swagger primitive type in dict form
     :type value: int, string, float, long, etc
     """
-    Swagger20RequestValidator(spec).validate(value)
+    Swagger20Validator(spec).validate(value)
 
 
 def validate_array(spec, value):
@@ -41,7 +41,7 @@ def validate_array(spec, value):
     :param spec: spec for an 'array' type in dict form
     :type value: list
     """
-    Swagger20RequestValidator(spec).validate(value)
+    Swagger20Validator(spec).validate(value)
 
 
 def validate_object(spec, value):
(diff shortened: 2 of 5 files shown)


bravado/mapping/validate.py
```diff
@@ -50,7 +41,7 @@ def validate_array(spec, value):
     :param spec: spec for an 'array' type in dict form
     :type value: list
     """
-    jsonschema.validate(value, spec)
+    Swagger20RequestValidator(spec).validate(value)
```
dnephin: This is a great change because it won't call `check_schema()` any more as well, which ends up being relatively expensive.

dnephin:
This looks good to me, but I'd wait for a ship it from spatel before merging

bravado/mapping/swagger20_validator.py
```diff
@@ -0,0 +1,39 @@
+from jsonschema import validators, _validators
+from jsonschema.exceptions import ValidationError
+from jsonschema.validators import Draft4Validator
+
+
```
analogue: Some docs about how/why we're doing this would be good.

bravado/mapping/swagger20_validator.py
```diff
@@ -0,0 +1,39 @@
...
+    """A validator which performs no validation. Used to `ignore` some schema
+    fields during validation.
+    """
+    return
+
+
+def required_validator(validator, req, instance, schema):
+    """Swagger 2.0 expects `required` to be a bool in the Parameter object,
+    but a list of properties everywhere else.
+    """
```
analogue: full docstrings please

bravado/mapping/swagger20_validator.py
```diff
@@ -0,0 +1,39 @@
...
+    # TODO: Find a better way to identify if it is a Parameter schema
+    if schema.get('in'):
```
analogue: This would probably benefit from being moved to `bravado.mapping.spec.is_param_spec(..)` where similar friends live.

bravado/mapping/swagger20_validator.py
```diff
@@ -0,0 +1,39 @@
...
+        if req is True and not instance:
+            return [ValidationError("%s is required" % schema['name'])]
+    else:
+        return _validators.required_draft4(validator, req, instance, schema)
```
analogue: I'm a little concerned about the use of the non-public module `jsonschema._validators` here - is there anything to be concerned about?

bravado/mapping/swagger20_validator.py
```diff
@@ -0,0 +1,39 @@
...
+
+def required_validator(validator, req, instance, schema):
+    """Swagger 2.0 expects `required` to be a bool in the Parameter object,
+    but a list of properties everywhere else.
+    """
+    # TODO: Find a better way to identify if it is a Parameter schema
+    if schema.get('in'):
+        if req is True and not instance:
+            return [ValidationError("%s is required" % schema['name'])]
+    else:
```
analogue: necessary?

bravado/mapping/swagger20_validator.py
```diff
@@ -0,0 +1,39 @@
...
+def enum_validator(validator, enums, instance, schema):
+    """Swagger 2.0 allows enums to be validated against objects of type
+    arrays, like query parameter (collectionFormat: multi)
+    """
+    if schema.get('type') == 'array':
+        return (v for item in instance for v in _validators.enum(
+            validator, enums, item, schema))
+    return _validators.enum(validator, enums, instance, schema)
+
+Swagger20RequestValidator = validators.extend(
```
analogue: This validator is not restricted to requests only based on the call site in validate.py.   `SwaggerValidator` would suffice.

tests/mapping/swagger20_request_validator/enum_validator_test.py
```diff
@@ -0,0 +1,30 @@
...
+from bravado.mapping.swagger20_validator import enum_validator
+
+
+def test_multiple_jsonschem_call_if_enum_items_present_as_array():
+    enums = ['a1', 'b2', 'c3']
+    param_schema = {
+        'name': 'foo',
+        'in': 'query',
+        'type': 'array',
+        'items': {'type': 'array'},
```
analogue: Do you mean for this to be `'items': {'type': 'string'},`  ?  (even though not really relevant to the test)

bravado/mapping/swagger20_validator.py
dnephin: `SwaggerValidator` seems a bit too general, I think this is specifically for operations, right? so Maybe `SwaggerOperationValidator` ?

tests/mapping/swagger20_request_validator/enum_validator_test.py
```diff
@@ -0,0 +1,30 @@
...
+        'enum': enums
+        }
+    with mock.patch('jsonschema._validators.enum') as m:
+        [list(x) for x in enum_validator(
```
analogue: `for` loop preferred - http://stackoverflow.com/questions/14249355/python-list-comprehension-without-assignment

tests/mapping/swagger20_request_validator/enum_validator_test.py
```diff
@@ -0,0 +1,30 @@
...
+            None, enums, ['a1', 'd4'], param_schema)]
+    m.assert_any_call(None, enums, 'a1', param_schema)
+    m.assert_any_call(None, enums, 'd4', param_schema)
+
+
+def test_single_jsonschema_call_if_enum_instance_not_array():
+    enums = ['a1', 'b2', 'c3']
+    param_schema = {
```
analogue: This really should have a `type` and `in` for completeness.

bravado/mapping/swagger20_validator.py
```diff
@@ -0,0 +1,39 @@
...
+    return
+
+
+def required_validator(validator, req, instance, schema):
+    """Swagger 2.0 expects `required` to be a bool in the Parameter object,
+    but a list of properties everywhere else.
+    """
+    # TODO: Find a better way to identify if it is a Parameter schema
+    if schema.get('in'):
+        if req is True and not instance:
```
analogue: Should `not instance` be `instance is None` ?

tests/mapping/swagger20_request_validator/required_validator_test.py
```diff
@@ -0,0 +1,23 @@
...
+
+def test_fail_if_required_parameter_but_not_present():
+    param_schema = {'name': 'foo', 'in': 'query', 'required': True}
+    assert isinstance(required_validator(None, True, None, param_schema)[0],
+                      ValidationError)
+
+
+def test_pass_if_not_required_paramter_and_not_present():
+    param_schema = {'name': 'foo', 'in': 'query', 'required': True}
+    assert required_validator(None, False, None, param_schema) is None
```
analogue: I'm confused here - the method says "not required parameter" but the value of "required" is True.

bravado/mapping/swagger20_validator.py
prat0318: Ack. I will add a description. I would still want to separate it out from `validate.py` as `swagger20_validator` will grow much larger with many more validation customizations.

bravado/mapping/swagger20_validator.py
prat0318: Ack.

bravado/mapping/swagger20_validator.py
prat0318: Ack.

bravado/mapping/swagger20_validator.py
prat0318: Yeah, correct

bravado/mapping/swagger20_validator.py
prat0318: yeah, for the condition schema.get('in') is True but req is False, should not call the next validation.

bravado/mapping/swagger20_validator.py
prat0318: The author adheres towards backwards compatibility, above is a non-public method though. Other option is to copy the [lines](https://github.com/Julian/jsonschema/blob/e1c30237cf57a0c221ea5c9a2677ca2edf20b18e/jsonschema/_validators.py#L285-290) in a local method here, which should not be an issue as there no other associations to hidden methods/attributes inside of the method.

bravado/mapping/swagger20_validator.py
prat0318: Actually we use the same validator for response validation as well, so i vote for something like `Swagger20Validator` or so

tests/mapping/swagger20_request_validator/enum_validator_test.py
prat0318: yeah, will fix that.

tests/mapping/swagger20_request_validator/enum_validator_test.py
prat0318: Actually, they both were not needed in the previous test as well. So, removed them from the previous one.

tests/mapping/swagger20_request_validator/required_validator_test.py
prat0318: Fixed. Actually the value `required` was directly sent as the second parameter. Corrected now to pass the `required` field.

bravado/mapping/swagger20_validator.py
analogue: current impl sounds fine to me.

analogue:
Move the TODO and lgtm. Nice work!

Issue #134, opened 2015-03-29
Different Exception Type between Sync and Async client


laucia:
Swagger-py's HTTPError is an exact copy of request's one
https://github.com/Yelp/swagger-py/blob/master/swaggerpy/exception.py#L9-L21
https://github.com/kennethreitz/requests/blob/master/requests/exceptions.py#L13-L31

but the SynchronousHttpClient throws requests.exceptions.HTTPError, and AsynchronousHttpClient throws swaggerpy.exception.HTTPError which makes error handling closely tied to the client used.

replacing the HTTPError code in swaggerpy/exception.py by a from requests.exceptions import HTTPError could solve this and by backward compatible.

dnephin:
Duplicate of #100

prat0318:
Closing it. Progress is tracked from now in #100.

PR #137, opened 2015-04-01
Raise common HTTPError for both Sync and Async clients.

### LINKED ISSUE #100: Inconsistent exceptions
We've observed that the `SynchronousHttpClient` and `AsynchronousHttpClient` raise different exceptions  (`requests.exceptions.HTTPError` vs `swaggerpy.exception.HTTPError`).

We need to ensure that all exceptions at least have the same base class, regardless of client implementation.

We may be able to wait for sprockets, but we should still address this.

### LINKED ISSUE #108: handle_response_errors swallows stacktrace
```/swaggerpy/response.py#handle_response_errors``` is not preserving the stacktrace when reraising the errors, making debugging difficult. 

Moving to the 3 parameters version of raise would solve this :

```
        ....

exc_info = sys.exc_info()
raise sys.exc_info[0], e.args, sys.exc_info[2]
```

prat0318:
- Fixes #100 and #134 (duplicate)
- Fixes #108

### DIFF
--- a/swaggerpy/response.py
+++ b/swaggerpy/response.py
@@ -7,9 +7,11 @@
 """Code for checking the response from API. If correct, it proceeds to convert
 it into Python class types
 """
+import sys
+
 import swagger_type
 from swagger_type import SwaggerTypeCheck
-from swaggerpy.exception import CancelledError
+from swaggerpy.exception import CancelledError, HTTPError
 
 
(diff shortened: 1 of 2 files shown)


dnephin:
Not python3 compatible, but that's fine for now.

LGTM

PR #138, opened 2015-04-01
[bravado] Make common HTTP Error for Fido and Requests clients. Refer: #100

### LINKED ISSUE #100: Inconsistent exceptions

prat0318:
Similar to #137. Fixes #100 and #108 for bravado release.

### DIFF
--- a/bravado/requests_client.py
+++ b/bravado/requests_client.py
@@ -140,9 +140,8 @@ def add_response_detail_to_errors(e):
     :type e: :class: `requests.HTTPError`
     :raises HTTPError: :class: `bravado.exception.HTTPError`
     """
+    args = list(e.args)
     if hasattr(e, 'response') and hasattr(e.response, 'text'):
-        # e.args is a tuple, change to list for modifications
-        args = list(e.args)
         args[0] += (' : ' + e.response.text)
     raise HTTPError(*args), None, sys.exc_info()[2]
 
--- a/bravado/requests_client.py
+++ b/bravado/requests_client.py
@@ -6,10 +6,12 @@
 
 import logging
 import urlparse
+import sys
 
 import requests
 import requests.auth
 
+from bravado.exception import HTTPError
 from bravado.mapping.http_client import HttpClient
 from bravado.mapping.http_future import HttpFuture
 from bravado.mapping.response import ResponseLike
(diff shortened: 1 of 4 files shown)


bravado/requests_client.py
```diff
@@ -139,7 +141,7 @@ def add_response_detail_to_errors(e):
         args = list(e.args)
         args[0] += (' : ' + e.response.text)
         e.args = tuple(args)
-    raise e
+    raise HTTPError(*e.args), None, sys.exc_info()[2]
```
analogue: Can't you achieve the same thing without mutating `e`?

```python
raise HTTPError(*args), None, sys.exc_info()[2]
```

prat0318: Good point, i needed only when raising back e, but not now. I will add unit tests for expected returned type of e (the input `e` shoud be of type `requests.HTTPError` but not compulsory).

analogue: My wording could have been misinterpreted:
- unit test (implies testing the `if` conditional)
- docstrings because I didn't know the expected type of `e`

tests/requests_client/add_response_details_to_errors_test.py
```diff
@@ -0,0 +1,16 @@
+from mock import Mock
```
analogue: python filename: s/details/detail

bravado/requests_client.py
```diff
@@ -133,13 +135,16 @@ def add_response_detail_to_errors(e):
...
+    :type e: :class: `requests.HTTPError`
+    :raises HTTPError: :class: `bravado.exception.HTTPError`
     """
     if hasattr(e, 'response') and hasattr(e.response, 'text'):
         # e.args is a tuple, change to list for modifications
         args = list(e.args)
         args[0] += (' : ' + e.response.text)
-        e.args = tuple(args)
-    raise e
+    raise HTTPError(*args), None, sys.exc_info()[2]
```
analogue: I don't think `args` will have a value when the `if` stmt is `False`.  Good reason to tests both sides of the conditional with a unit test.

tests/requests_client/add_response_details_to_errors_test.py
prat0318: ack.

bravado/requests_client.py
prat0318: Yeah, correct. Will add another test as well.

analogue:
ship it!

PR #139, opened 2015-04-01
Set model attribute to None, instead of creating a default value.

### LINKED ISSUE #133: Remove the model attributes if they are not required and not present in the response
Instead of providing a default value, remove the model field if not present in the response.

Changes should be around [here](https://github.com/Yelp/swagger-py/blob/8b8cadb81494dec71a4896cb287382d690430b74/swaggerpy/swagger_model.py#L360)

prat0318:
- Fixes #133 
- Note: Even for an array attribute, it will now set it to `None` instead of an earlier empty list.

### DIFF
--- a/tests/resource_model_test.py
+++ b/tests/resource_model_test.py
@@ -157,7 +157,7 @@ class ResourceTest(unittest.TestCase):
             u'http://localhost/api-docs').api_test
         User = resource.testHTTP._models['User']
         self.assertEqual(
-            {"schools": [], "id": 0L, "date": None, "datetime": None},
+            {"schools": None, "id": None, "date": None, "datetime": None},
             User().__dict__
         )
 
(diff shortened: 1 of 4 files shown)


tests/resource_model_test.py
```diff
@@ -214,9 +214,9 @@ def test_types_of_model_attributes(self):
         resource = SwaggerClient.from_url(
             u'http://localhost/api-docs').api_test
         models = resource.testHTTP._models
-        user = models['User']()
-        school = models['School']()
-        self.assertTrue(isinstance(user.id, long))
+        school = models['School'](name='foo')
+        user = models['User'](id=1, schools=[school])
+        self.assertTrue(isinstance(user.id, int))
```
dnephin: why is this now an int and not a long?

prat0318: I have changed it back to long. And changed the `id` value to `1l`.

dnephin:
lgtm

PR #140, opened 2015-04-07
Add a RequestLike interface for unmarshaling incoming requests on the server side

analogue:
New:
- `bravado.mapping.request.RequestLike`
- Interface for unmarshaling server side requests - `bravado.mapping.request.unmarshal_request(...)` 
- Unmashaling individual Swagger request params - `bravado.mapping.param.unmarshal_param(...)`


bravado/mapping/operation.py
```diff
@@ -155,59 +153,7 @@ def response_callback(response_adapter):
         return self.swagger_spec.http_client.request(request, response_callback)
 
 
-def unmarshal_response(response, op):
```
analogue: This just moved to `bravado.mapping.response` with no change in functionality.

bravado/mapping/operation.py
```diff
@@ -155,59 +153,7 @@ def response_callback(response_adapter):
...
-    content_spec = response_spec['schema']
-    content_value = response.json()
-    if op.swagger_spec.config['validate_responses']:
-        validate_schema_object(content_spec, content_value)
-    result = unmarshal_schema_object(
-        op.swagger_spec, content_spec, content_value)
-    return response.status_code, result
-
-
-def get_response_spec(status_code, op):
```
analogue: This just moved to bravado.mapping.response with no change in functionality.

bravado/mapping/request.py
```diff
@@ -0,0 +1,60 @@
+from bravado.mapping.operation import log
+from bravado.mapping.param import unmarshal_param
+
+
+class RequestLike(object):
+    """
+    Define a common interface for bravado to interface with server side
+    request objects.
```
dnephin: "interface" is here twice

bravado/mapping/request.py
```diff
@@ -0,0 +1,60 @@
...
+
+    Subclasses are responsible for providing attrs for __required_attrs__.
+    """
+    __required_attrs__ = [
+        'path',     # dict of URL path parameters
+        'params',   # dict of parameters from the query string and request body.
```
dnephin: This shouldn't include anything from the request body, I think the comment is misleading.

I think this should be `query` to be consistent with the rest of bravado and the swagger spec

bravado/mapping/param.py
```diff
@@ -128,6 +132,73 @@ def marshal_param(param, value, request):
...
+    elif location == 'body':
+        # TODO: verify content-type header
+        raw_value = request.json()
+    else:
+        raise SwaggerMappingError(
+            "Don't know how to unmarshal_param with location {0}".
+            format(location))
+
+    if param_spec['type'] == 'array' and location != 'body':
+        raw_value = unmarshal_collection_format(param_spec, raw_value)
```
dnephin: I think there might be a missing test case for this array case? How would the above if/elif block work if the type was an array?

bravado/mapping/param.py
```diff
@@ -128,6 +132,73 @@ def marshal_param(param, value, request):
...
+            raw_value = request.params.get(param.name, None)
+        else:
+            raw_value = cast_param(request.params.get(param.name, None))
+    elif location == 'body':
+        # TODO: verify content-type header
+        raw_value = request.json()
+    else:
+        raise SwaggerMappingError(
+            "Don't know how to unmarshal_param with location {0}".
+            format(location))
```
dnephin: This if/elif block looks like a good candidate for a `get_casted_value()` function.

`raw_value` seems like a weird name, the "raw" value would be the value over the wire, but the contents of this variable are actually converted to a type.

bravado/mapping/param.py
```diff
@@ -128,6 +132,73 @@ def marshal_param(param, value, request):
...
+
+    :type param: :class:`bravado.mapping.param.Param`
+    :type request: :class:`bravado.mapping.request.RequestLike`
+    """
+    param_spec = get_param_type_spec(param)
+    location = param.location
+    cast_param = partial(cast_request_param, param_spec['type'], param.name)
+
+    if location == 'path':
+        raw_value = cast_param(request.path.get(param.name, None))
```
prat0318: May be i am missing something, why do we need to cast each param from `request_object`? When creating a request `client.getPetById(petId=42)` `petId` should already be in its correct form? When does it convert to `string`?

bravado/mapping/param.py
```diff
@@ -128,6 +132,73 @@ def marshal_param(param, value, request):
             format(location))
```
prat0318: A question on new version's L128 above, for `json.dumps` it would expect a `dict` as a param. So, will this be no longer valid now: `client.pet.addPet(body=Pet(id=1, name='bla')).result()` ?

bravado/mapping/param.py
prat0318: Or is this from the perspective of server-side?

bravado/mapping/param.py
```diff
@@ -128,6 +132,73 @@ def marshal_param(param, value, request):
...
+    validate_schema_object(param_spec, raw_value)
+    value = unmarshal_schema_object(param.swagger_spec, param_spec, raw_value)
+    return value
+
+
+CAST_TYPE_TO_FUNC = {
+    'integer': int,
+    'number': float,
+    'boolean': bool
+}
```
prat0318: Should also include `string` : `str`.

bravado/mapping/param.py
```diff
@@ -128,6 +132,73 @@ def marshal_param(param, value, request):
...
+    :param param_name: param name
+    :type  param_name: string
+    :param param_value: param value
+    :type  param_value: string
+    """
+    if param_value is None:
+        return None
+
+    try:
+        return CAST_TYPE_TO_FUNC.get(param_type, lambda x: x)(param_value)
```
prat0318: I don't think of any type which can be used just as `lambda x: x`. How about throwing a keyError instead `CAST_TYPE_TO_FUNC[param_type]`?

bravado/mapping/request.py
analogue: This was initially `query` but after going through the `pyramid.request.Request` docs, I made it more generalized because of the need to access form data (in the body) and query params.  

This is actually copied verbatim from Pyramid docs @ http://docs.pylonsproject.org/docs/pyramid/en/latest/api/request.html#pyramid.request.Request.params

The only other reasonable access points that I could see where more specific  `request.GET` and `request.POST

bravado/mapping/param.py
analogue: The key here is `location != body`.  When an array is in the body, there is no need for additional packing/unpacking (based on a separator) from a string value that originated from the request path, query, or form data .  Can you be more specific about which test case is missing?

bravado/mapping/param.py
analogue: `raw_value` is primarily used so that the use of `value` on line 168 is not overloaded.

bravado/mapping/param.py
analogue: Yes, this is server side - `unmarshal` implies "wire -> python" and `param` implies request object. The only place that a request parameter is unmarshaled is on the server side.

To address your original question, a 3rd party server side request object doesn't usually know the types of the values that are extracted from the request path, query, form data, or headers. Hence, the type is implicitly a string. The casting is necessary to coerce the value into a type that matches the swagger spec.  If the casting fails, that is totally ok...the validator will catch it downstream.

bravado/mapping/param.py
analogue: The new version's L107 marshals the `Pet` model object into a dict.  By the time it gets to L128, it is a dict.  However, the parameter to `json.dumps` need not always be a dict.  int, bool, string, etc are totally legit json objects.

bravado/mapping/param.py
analogue: The param_value is expected to be of type string (see previous comment about values being a string implicitly on server side request objects). Hence, there is no need to do any casting and `lambda x:x` just returns the param_value  unchanged.

bravado/mapping/param.py
analogue: 'lambda x:x' is effectively a no-op and totally expected behavior.

bravado/mapping/request.py
dnephin: This shouldn't reflect pyramid in any way. This interface is a bravado-core interface. We can use an adapter to make the pyramid request fit what bravado needs, like https://github.com/striglia/pyramid_swagger/blob/fe3b87706673c58929e2a2b0a9d3236f4e4517ca/pyramid_swagger/tween.py#L132

I actually think we can use that exact adapter implementation.

The adapter would use `request.GET`. You don't need to know the request method, because post always uses body which will be swagger param type of `body` or `form`.

bravado/mapping/request.py
```diff
@@ -0,0 +1,59 @@
...
+                'for `{1}`'.format(type(self), name))
+        raise AttributeError(
+            "'{0}' object has no attribute '{1}'".format(type(self), name))
+
+    def json(self, **kwargs):
+        """
+        :return: request content in a json-like form
+        :rtype: int, float, double, string, unicode, list, dict
+        """
+        raise NotImplementedError("Implement json() in {0}".format(type(self)))
```
dnephin: Why is this `json()` a special case and not just part of `__required_attrs__`. I think this should be just a regular attribute, and the adapter should handle the conversion.

bravado/mapping/param.py
dnephin: Where the array is not in the body. Earlier statements would have tried to cast the value to an int or float.

`location='query', param_spec['type'] == 'array'`

actually, I think I see what's happening now. if type is array, it goes through the whole casting process, but it's just a no-op. This is kind of confusing. I think this should be inlined into the original if/elif block so it's clear this is a distinct case, not an extension of the cast.

bravado/mapping/request.py
analogue: What about the remaining http methods other than GET and POST?

dnephin: http://docs.pylonsproject.org/docs/pyramid/en/latest/api/request.html#pyramid.request.Request.GET

`request.GET` has all query params, regardless of http method

dnephin: implementation is here: https://github.com/Pylons/webob/blob/master/webob/request.py#L818

bravado/mapping/request.py
analogue: It is consistent with `ResponseLike` and leaves open the flexibility to pass in args ala `json.loads`. It is functional both ways and not really worth nit-picking over.

bravado/mapping/request.py
analogue: Well if that isn't misleading, I don't know what is.

bravado/mapping/request.py
dnephin: It seems to be inconsistent though. Using the adapter approach, we could customize json deserialization in the adapter. 

Is there any way to actually pass in these kwargs using bravado-core?

swagger calls this "body" type, it would be good to be consistent with swagger terminology.

bravado/mapping/param.py
analogue: It is a tricky list of conditionals to handle.  The orignial if/else is purposely only testing `location` for this reason.

bravado/mapping/request.py
analogue: YMMV, not a big deal IMO.  `json()` returns a json like object.  I'd be open to changing this at some later date but it works for now.

PR #144, opened 2015-05-11
Make API compatible with swagger-py

analogue:
- Moved `HttpFuture` from bravado-core to bravado
- HTTPError raised on client or server side errors based on HTTP status code to match swagger-py
- See sister PR in bravado-core - https://github.com/Yelp/bravado-core/pull/12

### DIFF
--- a/bravado/http_future.py
+++ b/bravado/http_future.py
@@ -4,7 +4,6 @@ from bravado.exception import HTTPError
 DEFAULT_TIMEOUT_S = 5.0
 
 
-# TODO: move to bravado
 class HttpFuture(object):
     """A future which inputs HTTP params"""
 
--- a/bravado/http_future.py
+++ b/bravado/http_future.py
@@ -1,0 +1,68 @@
+# -*- coding: utf-8 -*-
+from bravado.exception import HTTPError
+
+DEFAULT_TIMEOUT_S = 5.0
+
+
+# TODO: move to bravado
+class HttpFuture(object):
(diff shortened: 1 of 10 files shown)


bravado/http_future.py
```diff
@@ -0,0 +1,67 @@
...
+        to receive the response
+        :return: swagger response return value when given a callback or the
+            http_response otherwise.
+        """
+        http_response = self.response_adapter(
+            self.future.result(timeout=timeout))
+
+        if self.response_callback:
+            swagger_return_value = self.response_callback(http_response)
+            return raise_http_error_based_on_status(
```
prat0318: from the method name, it looks like it should work like requests's `raise_for_status` i.e. return None if sucess else raise error. How about returning the `swagger_return_value` after calling `raise_http_error_based_on_status`?

prat0318:
lgtm... other than one issue listed above. Also, is `HttpFuture` moved back to `bravado` so that `response_callback` can be wrapped up with `raise_status` or there was some other reason for it?

bravado/http_future.py
analogue: ack

analogue:
@prat0318 In general, bravado-core needs to get out of the business of making http calls. This was hard to do in the first cut of the refactoring because `bravado_core.spec.Spec` is dom-like in nature and the http calls are tied to a dom-like object - `bravado_core.operation.Operation` which is reached via traversal.

dnephin:
lgtm

PR #145, opened 2015-05-13
Bravado ready for RC1

analogue:
This includes all the work on the swagger2.0 branch plus the following in the last three commits:

- Updates to the swagger2.0 branch to make the merge green
- Updated sphinx documentation for bravado
- Bumped version to 0.1.0-rc1


CHANGELOG.rst
```diff
@@ -1,50 +1,3 @@
...
-
-0.2.0 (2013-10-28)
-------------------
-- Add close() methods to client and http_client.
-
-0.1.0 (2013-10-24)
-------------------
-
-- Initial release
+0.1.0-rc1 (2015-05-13)
```
dnephin: Just one question about versions.

The git repo currently has tags for all of our swagger-py release, so if we start again at 0.1, we're going to overlap.

It would probably be good to keep the tags instead of erasing them entirely, because we do still have people using the old versions for now at least.

Should we maybe call this 1.0.0-rc1 ?

analogue: Good point. I'll make that change.

dnephin:
lgtm

PR #149, opened 2015-05-28
Rename ResponseLike to IncomingResponse (due to renaming in bravado-core)

analogue:
Changes required for https://github.com/Yelp/bravado-core/pull/23. Build will fail in travis until bravado-core version is bumped + released.


dnephin:
LGTM (once build is green)

PR #150, opened 2015-06-04
Add acceptance test to check allowance of extra fields in the resposne

### LINKED ISSUE #111: Add acceptance test for "accept unknown fields in response"
If the client receives extra data in the response, it should not raise an error.  A service should be able to add extra fields to its response object without having to do a api version bump, as long as the old fields are unmodified.

This has the potential to create a lot of failures. I don't believe there is a test for this in `swagger-py`, but we should verify it works correct in `bravado` before release.

prat0318:
Fixes #111


tests/client_test.py
```diff
@@ -254,5 +258,60 @@ def setUp(self):
...
-if __name__ == '__main__':
-    unittest.main()
+class MockHttpClient(object):
+
+    def __init__(self, server_response):
+        self.response = server_response
+
+    def request(self, request, response_callback):
+        return HttpFuture(
+            mock.Mock(),
```
analogue: Please make the type of the object that the Mock is mocking obvious using `spec`.

tests/client_test.py
```diff
@@ -254,5 +258,60 @@ def setUp(self):
...
+            mock.Mock(return_value=mock.Mock(
```
analogue: Same here.

tests/client_test.py
```diff
@@ -254,5 +258,60 @@ def setUp(self):
...
+                        "type": "integer",
+                    }
+                }
+            }
+        }
+    }
+    return SwaggerClient(Spec.from_dict(spec_dict, 'http://example.com/'))
+
+
+def test_client_ignores_unexpecte_response_fields(swagger_client):
```
analogue: s/unexpecte/unexpected

tests/client_test.py
```diff
@@ -254,5 +258,60 @@ def setUp(self):
...
+    swagger_client.swagger_spec.http_client = MockHttpClient({
+        "id": 12345,
+        "alias": "thing",
+    })
+    response = swagger_client.pet.getPet()
+    assert response.result() == dict(id=12345)
```
analogue: I would actually expect this assertion to fail. The result should contain the `alias` key.

prat0318: No, the main point of this test is actually to check that the extra key `alias` is ignored in the response by bravado (as `alias` isnt mentioned in the spec).

analogue: Maybe "ignored" could use some clarification here.  It is not ignored by bravado because `additionalProperties: true` is implicit when not specified. Hence, bravado is passing through the `alias` key to the client and it should be in the dict returned by `response.result()`. 

You can find a branch that verifies this behavior here: https://github.com/analogue/bravado/tree/spatel_test_unknown_fields

The test failure output:
```
=================================================================== FAILURES ===================================================================
________________________________________________ test_client_ignores_unexpecte_response_fields _________________________________________________

swagger_client = SwaggerClient(http://example.com/)

    def test_client_ignores_unexpecte_response_fields(swagger_client):
        # swagger_client.swagger_spec.http_client = MockHttpClient({
        #     "id": 12345,
        #     "alias": "thing",
        # })
        response = swagger_client.pet.getPet()
>       assert response.result() == dict(id=12345)
E       assert {'alias': 'th..., 'id': 12345} == {'id': 12345}
E         Omitting 1 identical items, use -v to show
E         Left contains more items:
E         {'alias': 'thing'}
E         Use -v to get the full diff

tests/client_test.py:330: AssertionError
```

prat0318: Ah, that's a good point and worth adding as a comment in the test and to the README as well. As this PR is of low priority i will pick this up later.

analogue:
Don't think there is any more to do here. Working as designed.

PR #152, opened 2015-06-09
py3 support for bravado

bpicolo:
Should be all we need for py34.

The downside is it's not yet possible to use fido with py3 (twisted doesn't fully support it yet). So, the fido client raises an import error when you try to use it in py3.

I'm going to try to figure out a suitable async client in the meantime for py3 (though probably would make it a separate package or something of that nature)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: python
 env:
 - TOX_ENV=py26
 - TOX_ENV=py27
+- TOX_ENV=py34
 - TOX_ENV=cover
 - TOX_ENV=docs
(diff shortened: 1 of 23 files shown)


dnephin:
We were talking about using https://github.com/KeepSafe/aiohttp to make an async client, right?  I forget if there were any issues with that. It seems like a good option

dnephin:
lgtm

bpicolo:
The parts of yelp uri we use dont seem to have py3 issues but I will try to

prat0318:
Merged :+1:

PR #157, opened 2015-07-07
Pass responses appropriately to errors

### LINKED ISSUE #156: bravado should be able to parse errors as json
Along with the idea of uniform exceptions, we definitely expect errors to come back with JSON responses in, well, most all occasions in our own apis.

Bravado should be able to parse exception messages that are json. Otherwise taking any action on them is virtually impossible.

Here's an example from (swaggerpy):

(Pdb) e
HTTPError(u'403 Client Error: Forbidden : {\n    "error": {\n        "code": "FLAGGING_FORBIDDEN", \n        "description": "You can\'t flag this."\n    }\n}',)
(Pdb) e.message
u'403 Client Error: Forbidden : {\n    "error": {\n        "code": "FLAGGING_FORBIDDEN", \n        "description": "You can\'t flag this."\n    }\n}'

I can't even json.loads it myself because we stuff the status code in there. : (

bpicolo:
Fixes #156 , the main issue being that we actually weren't passing responses to exceptions properly.

e.json might make a good convenience method, but I skipped it for now (e.response.json() is a fine substitute. But documenting this at some point would be good)

### DIFF
--- a/swaggerpy/response.py
+++ b/swaggerpy/response.py
@@ -26,9 +26,13 @@ def handle_response_errors(e):
     :raises HTTPError: :class: `swaggerpy.exception.HTTPError`
     """
     args = list(e.args)
+    kwargs = {}
     if hasattr(e, 'response') and hasattr(e.response, 'text'):
         args[0] += (' : ' + e.response.text)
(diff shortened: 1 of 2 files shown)


bpicolo:
Fixed base branch

PR #158, opened 2015-07-07
Pass responses appropriately to errors (swaggerpy)

### LINKED ISSUE #156: bravado should be able to parse errors as json

bpicolo:
Fixes #156 , the main issue being that we actually weren't passing responses to exceptions properly.

e.json might make a good convenience method, but I skipped it for now (e.response.json is a good substitute. But documenting this at some point would be good)

### DIFF
--- a/swaggerpy/response.py
+++ b/swaggerpy/response.py
@@ -26,9 +26,17 @@ def handle_response_errors(e):
     :raises HTTPError: :class: `swaggerpy.exception.HTTPError`
     """
     args = list(e.args)
-    if hasattr(e, 'response') and hasattr(e.response, 'text'):
-        args[0] += (' : ' + e.response.text)
-    raise HTTPError(*args), None, sys.exc_info()[2]
+    kwargs = {}
+
+    if hasattr(e, 'response'):
+        kwargs['response'] = e.response
+        if hasattr(e.response, 'text'):
+            args[0] += (' : ' + e.response.text)
+
+    if hasattr(e, 'request'):
+        kwargs['request'] = e.request
+
+    raise HTTPError(*args, **kwargs), None, sys.exc_info()[2]
 
 
 class HTTPFuture(object):
--- a/tests/resource_response_test.py
+++ b/tests/resource_response_test.py
@@ -55,16 +55,17 @@ is validated against its type 'Pet' which is defined like so:
 """
 
 import datetime
-from swaggerpy.compat import json
 import unittest
-from mock import Mock
+from swaggerpy.compat import json
 
 import httpretty
 from dateutil.tz import tzutc
+from mock import Mock
 
 from swaggerpy.client import SwaggerClient
 from swaggerpy.exception import CancelledError, HTTPError
 from swaggerpy.processors import SwaggerError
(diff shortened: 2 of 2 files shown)


swaggerpy/response.py
```diff
@@ -26,9 +26,13 @@ def handle_response_errors(e):
     :raises HTTPError: :class: `swaggerpy.exception.HTTPError`
     """
     args = list(e.args)
+    kwargs = {}
     if hasattr(e, 'response') and hasattr(e.response, 'text'):
```
analogue: Is there value in setting `kwargs['response']` even though `e.response.text` doesn't exist?

bpicolo: ah, yeah, definitely. fixed

tests/resource_response_test.py
```diff
@@ -87,6 +88,22 @@ def test_cancelled_returns_false_if_called_before_cancel(self):
...
+    def test_pass_response_and_request(self):
+        request = Mock()
+        response = Mock(text='{"foo": "bar"}')
+        error = Mock(
+            args=('400 Bad Request',), response=response, request=request
+        )
+        try:
+            handle_response_errors(error)
+        except HTTPError as e:
+            self.assertEqual(e.json, {'foo': 'bar'})
```
analogue: Where is `e.json` getting set?

analogue:
lgtm

would be good to have @prat0318 give it a once over since he is more familiar with the codebase.

tests/resource_response_test.py
bpicolo: Ah, that was from previous diff where I had added that as a property to exception. Fixed <3

tests/resource_response_test.py
```diff
@@ -87,6 +88,23 @@ def test_cancelled_returns_false_if_called_before_cancel(self):
...
+        response.json.return_value = {'foo': 'bar'}
+        error = Mock(
+            args=('400 Bad Request',), response=response, request=request
+        )
+        try:
+            handle_response_errors(error)
+        except HTTPError as e:
+            self.assertEqual(e.response.json(), {'foo': 'bar'})
+            self.assertEqual(e.response, response)
+            self.assertEqual(e.request, request)
```
prat0318: I am not sure if there is a test already that '400 Bad Request' is contained in the `e`'s args, will be good to add it if not present already.

prat0318:
lgtm

PR #159, opened 2015-07-10
Move http invocation to bravado

analogue:
Related PR on the bravado-core side - https://github.com/Yelp/bravado-core/pull/31

This just wraps the `Resource` and `Operation` objects from `bravado-core` to allow invocation of the http call.

The `operation_docstring_wrapper` is a weird workaround for the (only?) way to get docstrings generated  and for the operation to still be callable.

Tested with a local client and works as expected.


dnephin:
lgtm

PR #163, opened 2015-07-16
Add log warning for deprecated op

### LINKED ISSUE #162: log.warn `deprecated`, `x-deprecated-date` and `x-removal-date`

prat0318:
Fixes #162


prat0318:
@analogue @dnephin whenever you have time ^

bravado/client.py
```diff
@@ -215,6 +215,16 @@ def construct_params(self, request, op_kwargs):
...
                 marshal_param(remaining_param, None, request)
 
+    def log_warning_for_deprecated_op(self):
+        # TODO: Move this logic to bravado_decorators
+        if self.operation.op_spec.get('deprecated'):
+            params = {'op': self.operation.operation_id,
+                      'dep': self.operation.op_spec.get('x-deprecated-date'),
+                      'rem': self.operation.op_spec.get('x-removal-date'),
+                      }
+            log.warn(("[DEPRECATED] {op} has now been deprecated. Deprecation"
```
prat0318: As per [SO answer](http://stackoverflow.com/a/14762106) should this be changed to `warnings.warn`?

bravado/client.py
```diff
@@ -215,6 +215,16 @@ def construct_params(self, request, op_kwargs):
             if not remaining_param.required and remaining_param.has_default():
                 marshal_param(remaining_param, None, request)
 
+    def log_warning_for_deprecated_op(self):
+        # TODO: Move this logic to bravado_decorators
+        if self.operation.op_spec.get('deprecated'):
```
analogue: While functionally equivalent, specifying a default value of `False` here is a little more explicit and in line with the [spec](https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#operation-object)
```python
if self.operation.op_spec.get('deprecated', False):
```

bravado/client.py
```diff
@@ -215,6 +215,16 @@ def construct_params(self, request, op_kwargs):
             if not remaining_param.required and remaining_param.has_default():
                 marshal_param(remaining_param, None, request)
 
+    def log_warning_for_deprecated_op(self):
+        # TODO: Move this logic to bravado_decorators
```
analogue: I think this is actually fine remaining in bravado.  Treating `x-deprecated-date` and `x-removal-date

bravado/client.py
analogue: This would benefit from using python's `warnings` module.

tests/client/OperationDecorator/log_warning_for_deprecated_op_test.py
```diff
@@ -0,0 +1,39 @@
...
+from bravado_core.operation import Operation
+from bravado_core.spec import Spec
+
+from bravado.client import OperationDecorator
+
+
+@pytest.fixture
+def getPetById_spec(petstore_dict):
+    op = petstore_dict['paths']['/pet/{petId}']['get'].copy()
+    op.update({'deprecated': True,
```
analogue: Additional test cases I can think of:
- test_deprecated_false()
- test_deprecated_not_present()

prat0318: ack

bravado/client.py
prat0318: ack

bravado/client.py
prat0318: Can you tell more optional fields, you mean those two can be optional fields to this method?

bravado/client.py
prat0318: I had it in mind, but then went against it because of warnings logging to stderr. I am open to either.

prat0318:
Add PR review suggestions. ^

bravado/client.py
analogue: So `deprecated` is part of the Swagger spec, so we're obviously supporting that functionality.

However, `x-deprecated-date` and `x-removal-date` are probably only going to be used within Yelp.  Having to split this functionality out into a separate internal package is probably more hassle than it is worth for a couple of reasons. First, this isn't a core feature. Second, all it does is surface more detailed info in the deprecation warning.

As such, the code should treat the `x-` fields as optional.  If they exist, the deprecation message should be updated accordingly.  Each field stands on its own and does not require the other, I would imagine.

prat0318: As the fields are only used for displaying the warning message, i was reluctant to add all the if's to check the `None`ness of the fields and not displaying that field accordingly. I feel `x-foo: None` in the message conveys that it is not provided.

I would also vote to move this method to some other module, as it is not super related to `client`.

analogue: Thats cool with me.

prat0318:
Moved warning logic to a separate module

bravado/warning.py
```diff
@@ -0,0 +1,22 @@
...
+
+        dep_date = op.op_spec.get('x-deprecated-date')
+        if dep_date:
+            message += "Deprecation Date: {0}. ".format(dep_date)
+
+        rem_date = op.op_spec.get('x-removal-date')
+        if rem_date:
+            message += "Removal Date: {0}".format(rem_date)
+
+        warnings.warn(message, Warning)
```
bpicolo: If you want to  be super explicit, s/Warning/DeprecationWarning

bravado/warning.py
```diff
@@ -0,0 +1,22 @@
+import warnings
+
+
+def warn_for_deprecated_op(op):
+    """Warn if requested operation has `deprected` field flagged as True
```
analogue: s/deprected/deprecated

analogue:
lgtm

bravado/warning.py
prat0318: ack

bravado/warning.py
prat0318: Used Warning instead of DeprecationWarning, as the latter is by default turned off and hence the warning will get missed.

Issue #165, opened 2015-07-22
timeout and connect_timeout are not being passed to the FidoClient


analogue:
The call to [fido.fetch(....)](https://github.com/Yelp/bravado/blob/3c00a50f8a7d634cf179d771d371a200a22c0df6/bravado/fido_client.py#L68) doesn't currently pass along the `timeout` kwarg. 

After https://github.com/Yelp/fido/issues/12 is fixed, it should also pass along the ```connect_timeout``` kwarg too.

This is not to be confused with the ```timeout``` passed to ```future.result(timeout)``` since time may have passed between to call to ```fido.fetch(..)``` and ```future.result(...)```.

Issue #160, opened 2015-07-14
swagger client should not be modified by a service call.


podarsmarty:
For swaggerpy version 0.7.6 when we call and endpoint with a body parameter the header {'content-type': 'application/json'} is added for us. However, it is persisted which causes endpoints that don't need that header to fail.

i.e. -
client.get_client(...).endpoints.endpoint(body={'some_data': 'some_data'})

workaround for now is

client.get_client(...).endpoints.endpoint(_request_options={'headers': {'content-type': 'application/json'}}, body={'some_data': 'some_data'})

prat0318:
So this issue is that [here](https://github.com/Yelp/bravado/blob/master/bravado/client.py#L180) we are fetching the `_request_options` reference and [mutating](https://github.com/Yelp/bravado/blob/master/bravado/client.py#L186) [it](https://github.com/Yelp/bravado/blob/swaggerpy/swaggerpy/client.py#L491) later. The dict provided by user should not get mutated at any cost. So, i think the fix should be:
```python
_request_options = kwargs.pop('_request_options', {}).copy() or {}
```
which will needed to be patched to both `swaggerpy` and `bravado`.

lucafavatella:
Do I understand correctly this was fixed by #166?

PR #166, opened 2015-07-24
Support timeouts in _request_options

analogue:
So `_request_options` currently only supports `headers`. This adds support for `timeout` and `connect_timeout` to both the Fido and Requests http clients.

Fixes issue #165

### DIFF
--- a/bravado/client.py
+++ b/bravado/client.py
@@ -174,8 +174,8 @@ class OperationDecorator(object):
 
     def construct_request(self, **op_kwargs):
         """
-        :param op_kwargs: parameter name/value pairs to pass to the invocation
-            of the operation.
+        :param op_kwargs: parameter name/value pairs to passed to the
+            invocation of the operation.
         :return: request in dict form
         """
         request_options = op_kwargs.pop('_request_options', {})
@@ -183,16 +183,22 @@ class OperationDecorator(object):
         request = {
             'method': self.operation.http_method.upper(),
             'url': url,
-            'params': {},
+            'params': {},  # filled in downstream
             'headers': request_options.get('headers', {}),
         }
+
+        # Copy over optional request options
+        for request_option in ('connect_timeout', 'timeout'):
+            if request_option in request_options:
+                request[request_option] = request_options[request_option]
+
         self.construct_params(request, op_kwargs)
         return request
 
     def construct_params(self, request, op_kwargs):
         """
         Given the parameters passed to the operation invocation, validates and
-        marshals the parmameters into the provided request dict.
+        marshals the parameters into the provided request dict.
 
         :type request: dict
         :param op_kwargs: the kwargs passed to the operation invocation
@@ -225,10 +231,10 @@ class OperationDecorator(object):
         """
         log.debug(u"%s(%s)" % (self.operation.operation_id, op_kwargs))
         warn_for_deprecated_op(self.operation)
-        request = self.construct_request(**op_kwargs)
+        request_params = self.construct_request(**op_kwargs)
 
         def response_callback(response_adapter):
             return unmarshal_response(response_adapter, self)
 
         return self.operation.swagger_spec.http_client.request(
-            request, response_callback)
+            request_params, response_callback)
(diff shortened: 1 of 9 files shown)


bravado/client.py
```diff
@@ -174,8 +174,8 @@ def __getattr__(self, name):
 
     def construct_request(self, **op_kwargs):
         """
-        :param op_kwargs: parameter name/value pairs to pass to the invocation
-            of the operation.
+        :param op_kwargs: parameter name/value pairs to passed to the
+            invocation of the operation.
         :return: request in dict form
         """
         request_options = op_kwargs.pop('_request_options', {})
```
bpicolo: Mind adding .copy() here while you're at it?

analogue: Will create another PR to make sure `_request_options` and `headers` are not mutated.

prat0318: This will fix #160 for bravado.

analogue: Doesn't look like it is an issue in bravado - https://github.com/Yelp/fido/blame/master/fido/fido.py#L171

bravado/requests_client.py
```diff
@@ -185,18 +218,55 @@ def check_for_exceptions(self, response):
...
+        """
+        Build the appropriate timeout object to pass to `session.send(...)`.
+
+        :param result_timeout: timeout that was passed into `result(..)`.
+        :return: timeout
+        :rtype: float or tuple(connect_timeout, timeout)
+        """
+        timeout = None
+        has_service_timeout = 'timeout' in self.misc_options
+        has_result_timeout = result_timeout is not None
```
prat0318: As result_timeout is being passed from [HttpFuture](https://github.com/analogue/bravado/blob/pass_timeouts/bravado/http_future.py#L32) `result_timeout` will always have some value (or a default value of 5.0), so `has_result_timeout` flag is always true.

bravado/requests_client.py
```diff
@@ -185,18 +218,55 @@ def check_for_exceptions(self, response):
...
+
+        service_timeout = self.misc_options.get('timeout')
+        if has_service_timeout and has_result_timeout:
```
prat0318: Can use `service_timeout` instead of `has_service_timeout`.

bravado/requests_client.py
```diff
@@ -93,6 +93,31 @@ def __init__(self):
         self.session = requests.Session()
         self.authenticator = None
 
+    @staticmethod
+    def separate_params(request_params):
+        """Splits the passed in dict of request_params into two buckets.
+
+        - sanitized_params are valid kwargs for constructing a
+          requests.Request(..)
+        - misc_options are things like timeouts which can't be communicated
```
prat0318: Can we use the variable name used inside as `misc_options` as well (instead of `request_options`) to make things clearer.

bravado/requests_client.py
analogue: Will the evaluation of L234 be the same if ```self.misc_options = {'timeout': 0}```?

bravado/requests_client.py
analogue: Is it true in this case?

```python
petstore.pet.getPetById(id=99).result(timeout=None)
```

bravado/requests_client.py
analogue: ack

bravado/requests_client.py
prat0318: ah.. right.

bravado/requests_client.py
prat0318: Well, that's fair point. But from `timeout`'s [docstring](http://bravado.readthedocs.org/en/latest/bravado.html#bravado.http_future.HttpFuture.result) behavior with `None` is undefined. I think with a default value provided, a user given `None` value is really a special case than some accepted scenario.

analogue: I actually think Fido shouldn't have a default here since it just turns around and calls [concurrent.futures.Future::result(timeout=None)](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Future.result) which does a good job of defining what passing `None` means (wait indefinitely).

Changing the default in fido is a breaking change which I'm not ready to take on atm.

bravado/requests_client.py
analogue: Granted, setting it to zero is pretty much useless but I do prefer the readability and consistency of the `has_*` usage in this method.

prat0318:
looks good overall.

PR #167, opened 2015-08-08
Unified exceptions and better exception handling

### LINKED ISSUE #153: Bravado should reraise exceptions into common exception types
Shouldn't have to worry about this across separate clients. Should have uniform exceptions

analogue:
Fixes #153 

```bravado.exception.HTTPError``` is raised consistently by both Requests and Fido clients. 
```bravado_core.exception.SwaggerMappingError``` is raised during swagger outgoing request processing or swagger incoming response processing failures.

The exception story:
- When there is no response_callback (just a plain old http req/resp) then ```HTTPError``` is raised
- When making a service call that fails with a 5XX, then ```HTTPError``` is raised with minimal info
- When making a service call that succeeds, no exception is raised
- When making a service call that has a non-2XX response that matches a response in the spec, a ```HTTPError``` is raised with the ```swagger_result``` noted in the spec
- When making a service call that has a non-2XX response that doesn't match any responses in the spec, then ```HTTPError``` is raised without a ```swagger_result```

Depends on https://github.com/Yelp/bravado-core/pull/39


bpicolo:
Lg2m : )

dnephin:
lgtm once travis is green

Issue #169, opened 2015-08-11
http_client by default is None


stratosgear:
Am I using this wrong?

    from bravado.client import SwaggerClient
    from bravado.swagger_model import load_file
    from bravado.requests_client import RequestsClient

    http_client = RequestsClient()

    client = SwaggerClient.from_spec(load_file("/server/plaavi.json"), http_client=http_client)

    jobs = client.jobs.getJobs().result()

If I do not specifically instantiate a ```RequestsClient``` (and pass it to SwaggerClient) I get an error:

    ...
    File "/usr/local/lib/python2.7/dist-packages/bravado_core/docstring.py", line 52, in wrapper
      return operation(**kwargs)
    File "/usr/local/lib/python2.7/dist-packages/bravado/client.py", line 241, in __call__
      return self.operation.swagger_spec.http_client.request(request_params,
    AttributeError: 'NoneType' object has no attribute 'request'

which kinda makes sense because I see that the method param is defined as ```http_client=None```.

I'm using version 4.0.0 of bravado, btw

Thanks!

PR #170, opened 2015-08-11
Set a default http_client if none used

stratosgear:
Maybe something like this would work better (regarding #169)


analogue:
:+1:

PR #175, opened 2015-10-12
Bump major version for upgrade to bravado-core 3.0.2

analogue:
- New user-defined format registration mechanism is not backwards compatible so bump major version
- Updated docs
- Added ``make docs``


docs/source/configuration.rst
```diff
@@ -1,10 +1,17 @@
...
+Configuration
+=============
 
 You can configure certain behaviours when creating a ``SwaggerClient``.
 
+The full documentation for each config key is in `bravado-core <http://bravado-core.readthedocs.org/en/latest/config.html>`_.
+
 .. code-block:: python
 
+    from bravado_core.formatter import SwaggerFormat
```
prat0318: I don't think we need to expose `bravado-core` interface to a `bravado` user. I would like to have it imported in `bravado.client` or `bravado.formatter` module, and let the user import that instead.

analogue: It is a pretty useless layer of indirection for not much gain but OK.

Issue #173, opened 2015-09-24
Add a way to fetch response headers, status_code


prat0318:
Currently, the only output back is either json response or deserialized version of the json response. There are use-case where other attributes of response may be needed like headers, status_code, etc. There should be a way to fetch these as well.

analogue:
Fixed in 6.1.0

PR #176, opened 2015-10-14
Make the return type of HttpFuture.result() configurable

analogue:
See #173 for context.

- Defaults to the current behavior to maintain backwards compatibility. 
- I will update the sphinx docs on how to use this feature in detail once the implementation is reviewed.

Configure to return an http response:
```python
client = SwaggerClient(..., config={'return_swagger_result':False})
http_response = client.pet.getPetById(petId=42).result()
assert isinstance(response, bravado_core.response.IncomingResponse)
pet = http_response.swagger_result
assert http_response.status_code == 200
```

Configure to return the swagger result (default):
```python
client = SwaggerClient(..., config={'return_swagger_result':True})
pet = client.pet.getPetById(petId=42).result()
assert pet.name
```

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,11 @@
 Changelog
 =========
 
+6.1.0 (2015-XX-XX)
+------------------
+- Clients can now access the HTTP response from a service call to access things
+  like headers and status code. See `Advanced Usage <http://http://bravado.readthedocs.org/en/latest/advanced.html#getting-access-to-the-http-response>`_
+
 6.0.0 (2015-10-12)
 ------------------
 - User-defined formats are no longer global. The registration mechanism has
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,7 +4,7 @@ Changelog
 6.1.0 (2015-XX-XX)
 ------------------
 - Clients can now access the HTTP response from a service call to access things
-  like headers and status code. See `Advanced Usage <http://http://bravado.readthedocs.org/en/latest/advanced.html#getting-access-to-the-http-response>`_
+  like headers and status code. See `Advanced Usage <http://bravado.readthedocs.org/en/latest/advanced.html#getting-access-to-the-http-response>`_
 
 6.0.0 (2015-10-12)
 ------------------
--- a/bravado/client.py
+++ b/bravado/client.py
@@ -69,7 +69,7 @@ log = logging.getLogger(__name__)
 CONFIG_DEFAULTS = {
     # See the constructor of :class:`bravado.http_future.HttpFuture` for an
     # in depth explanation of what this means.
-    'return_swagger_result': True
+    'also_return_response': False
 }
 
 
@@ -255,12 +255,12 @@ class CallableOperation(object):
         warn_for_deprecated_op(self.operation)
         request_params = self.construct_request(**op_kwargs)
         callback = functools.partial(response_callback, operation=self)
-        return_swagger_result = \
-            self.operation.swagger_spec.config['return_swagger_result']
+        also_return_response = \
+            self.operation.swagger_spec.config['also_return_response']
         return self.operation.swagger_spec.http_client.request(
             request_params,
             callback,
-            return_swagger_result=return_swagger_result)
+            also_return_response)
 
 
 def response_callback(incoming_response, operation):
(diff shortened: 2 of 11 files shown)


prat0318:
So, i would have to choose one among the two. :-/ Mostly, i see the use-case being getting the response (swaggerized) + say, status_code or response_headers for extra checking/validation i would like to do along. 

I know it is ideal either but how about attaching the `IncomingResponse` attrs to the response object through some new attribute `_response_properties`.

analogue:
Yeah, this is a hard change given the need to maintain backwards compatibility.  

Returning the http response is the most flexible IMO - it provides access to everything you could possibly need.

You can't attach anything to the `swagger result` because the result may be `None`.

prat0318:
To maintain BC, how about the option we provide is : `swagger_result_with_response` which remains by default as False. And when the use choses the option, we return both (may be like a tuple or something). I feel the option of chosing between a or a+b is better than a or b.

bpicolo:
The swagger result is just a callback. Just allow people to change the callback and it's as configurable as you want

analogue:
Well, the callback is not exactly meant to be "pluggable". I think it ended up that way to support both the http request to retrieve the spec (no callback needed) and subsequent http requests to make service calls (callback necessary). 

In the case of service calls, the callback has the responsibility of handing the incoming response off to bravado-core for validation and unmarshalling and then returning the swagger result.  This is behavior that we want to retain.

This also has a bit of overlapping concerns with another issue that came up during proposed changes to the zipkin client decorator (still need to create a ticket for that). Basically, we need to support a pipeline for post-processing the incoming response after bravado-core is done but before the swagger result is returned to the client. My initial idea is to support a list of callbacks instead of just a single one and each gets invoked in turn. Regardless, the very last step in the pipeline still needs to support a configurable return type (the choice here is binary, so exposing this as a callback may be overkill).

analogue:
@prat0318 I think the tuple `(swagger_result, http_response)` will work since it is easier to ignore the response if not needed.

prat0318:
you mean the tuple as the non-default option (with keeping BC)? - yeah i like that.

prat0318:
On second thoughts, can the config be instead be in `result()` configs, so that i can be controlled per request. From perspective of zipkin decorator, it won't be able to control the configs of `SwaggerClient` but can control `result` configs.

analogue:
This specific feature is client wide.

prat0318:
ok. but i suppose after this feature, it should not be too hard adding it request wise (request config should be overriding the client config). should i create a new issue for this?

analogue:
I've got a ticket drafted -- I'll create it shortly.  Per-request configuration is going to be messy. I've got some ideas about how to implement this so we can discuss further once I start work on the PR.

prat0318:
:+1: will review this PR up by eod.

CHANGELOG.rst
```diff
@@ -1,6 +1,11 @@
 Changelog
 =========
 
+6.1.0 (2015-XX-XX)
+------------------
+- Clients can now access the HTTP response from a service call to access things
+  like headers and status code. See `Advanced Usage <http://http://bravado.readthedocs.org/en/latest/advanced.html#getting-access-to-the-http-response>`_
```
prat0318: typo: http://http://

bravado/client.py
```diff
@@ -271,15 +282,16 @@ def response_callback(incoming_response, operation):
     raise_on_unexpected(incoming_response)
 
     try:
-        swagger_return_value = unmarshal_response(incoming_response, operation)
+        incoming_response.swagger_result = unmarshal_response(
```
prat0318: now that we are returning back a tuple having separate fields for swagger_result and response, we can do away with polluting `incoming_response` object with `swagger_result`?

analogue: I actually want to keep it this way so a single object acts as a gateway to all contained concerns. Similiar to the server side with the request object in a pyramid in a view:

```python
@view('blah')
def get_pet(request):
    pet_id = request.swagger_data['pet_id']
```
Also, will make it simpler to pass around just the response to callbacks in #178.

prat0318: i think we should then document that the result interface of `incoming_response` is not exactly the one defined in `bravado_core.response.IncomingResponse`, but has some extra attributes added.

analogue: Yeah, I'll document that as part of #178. It is not exactly relevant to the public interface in this PR.

prat0318: kk, will try to cover the rest of the PR today.

prat0318:
:ship:

Issue #130, opened 2015-03-24
Make `RequestsClient` immutable.


prat0318:
See the comments from #127

analogue:
This looks fixed in https://github.com/Yelp/bravado/pull/127

dnephin:
It's mostly done, but the authentication stuff is still mutable (and it doesn't need to be).

PR #180, opened 2015-10-18
Configurable ResourceDecorator

isysd:
I added an optional resource_decorator parameter to the client, which allows for custom authentication, among other things. I believe this is an adequate solution for #130.


isysd:
hmm I have some issues. Will clean up the extra files and resubmit.

Issue #178, opened 2015-10-15
Support for post-processing the incoming response


analogue:
Client decorators need the ability to do post-processing on the incoming response in the ``HttpFuture`` after it has been validated and unmarshalled by bravado-core but before it is returned to the client. Expose an interface to make this possible.

analogue:
Fixed in 7.0.0

PR #181, opened 2015-10-19
Configurable ResourceDecorator

isysd:
I added an optional resource_decorator parameter to the client, which allows for custom authentication, among other things. I believe this is an adequate solution for #130.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,11 @@
 Changelog
 =========
 
+6.1.0 (2015-XX-XX)
+------------------
+- Clients can now access the HTTP response from a service call to access things
+  like headers and status code. See `Advanced Usage <http://http://bravado.readthedocs.org/en/latest/advanced.html#getting-access-to-the-http-response>`_
+
 6.0.0 (2015-10-12)
 ------------------
 - User-defined formats are no longer global. The registration mechanism has
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,7 +4,7 @@ Changelog
 6.1.0 (2015-XX-XX)
 ------------------
 - Clients can now access the HTTP response from a service call to access things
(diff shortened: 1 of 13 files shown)


analogue:
If I understand your issue correctly, wrapping the client/resource/operation is preferable because it allows instrumentation at each layer of composition instead of limiting it to a single callback.

For example:

```python
class AuthenticatingClient(object):
    def __init__(self, delegate):
        self.delegate = delegate

    def __get__(self, name):
        """Customize resource"""
        # get resource from self.delegate and wrap 
        pass
```

isysd:
Exactly. We are using bravado with jws (JSON Web Signatures). Among other things, these messages have different content types, and some header info in the body. An experimental example of how we're using this change is available here:

https://github.com/deginner/bravado-bitjws

analogue:
^ unrelated changes - I must have done something on the wrong branch

isysd:
I merged the new commits from upstream master. They're unrelated, but doing this merge makes your job easier, no? Should I resubmit the pull request?

analogue:
Hold off on that. I took a brief look at bravado-bitjws and I think there is a much cleaner way of doing it without all the code duplication. I'll have to get back to you on it when I have some time.

I'm currently working on a related concern as part of #178 so I'll chime in again once I get a PR out for that.

isysd:
Thanks, it does seem like there should be a much cleaner way. Most of the duplication in bravado-bitjws is to provide easy import and initialization. The alternative is to construct all of the objects manually and build your way up to the SwaggerClient.

isysd:
After refining bravado-bitjws further, I was able to get rid of most of the duplication.

Still, this seems like this PR is an efficient way to go. I hard coded the ResourceDecorator for bravado-bitjws, but that is only for convenience.

isysd:
The callback options for v7.0.0 look like a good alternative. I will use those instead of modifying the ResourceDecorator, and close this PR.

PR #183, opened 2015-10-20
Experimental implementation for post-processing incoming responses

analogue:
See #178 for context.

@prat0318 See if this satisfies the zipkin decorator use case. I'll add unit tests and docs once the interface looks good.  It is unfortunate that we can't use the same decorator pattern we've been using for wrapping ``Client``, ``Resource`` and ``Operation`` since the point of instrumentation is internal to the ``HttpFuture``.

Reponse callbacks
```python
def foo(incoming_response, operation):
    print 'foo'

request_options = {
    'response_callbacks': [foo]    
}

client = Swagger.from_url(...)
pet = client.pet.getPet(petId=1234, _request_options=request_options).result()
```

**Edit**: Removed client-wide response callbacks since client decorators can insert additional callbacks in their respective operation decorator.


prat0318:
Few comments (i haven't gone through the implementation, so pardon me if there is something silly):
1) What will be the order of processing of callbacks if both client and per request are defined?
2) Why give a option of `list` for callbacks? How about just a single callback and let client figure out.
3) I think `incoming_response` and `operation` should be immutable when passed inside. Are they?
4) Overall, i feel if a similar `_request_options` option is provided for `also_return_response`, that will suffice the `zipkin` requirements. The callback is much more powerful (complex) and not needed for zipkin case.

analogue:
1) The order passed into ``_request_options``.  When using an operation decorator, the decorator can choose where to insert the callback into the existing list. I'd like to defer on using any ordering rules (like pyramids ingress/egress tween options) until/unless we have a use-case where it is necessary.

2) A list supports > 1 client decorator and the option to insert the callback at beginning/end of the list. I don't understand `How about just a single callback and let client figure out`. Can you give me an example?

3) Don't understand the question. How would you make ``incoming_response`` and ``operation`` immutable?

Unit tests have been added. 
4) Added ability to pass ``also_include_response`` via ``_request_options``.

prat0318:
1) I was more curious when both client and request callbacks are present. Now that client callback is removed, it is more clearer.

2) Something like:
```py
# collector defined by the user.
def collector(incoming_response, operation):
    for callback in mylist:
         callback(incoming_response, operation)

request_options = {'response_callback': collector}
```

3) I was thinking more like creating namedtuples out of the two and returning those to the callback. But then, i am not sure how useful the feature will remain (as that is already provided by _also_include_response).

4) ++

analogue:
2) In your example, you're putting the responsibility of collecting and executing the callbacks on the bravado client instead of leaving it up to each individual SwaggerClient decorator.  The responsibility is misplaced in this case. Imagine you have a chain of decorators:
```python
client = SwaggerClient.from_url(...)
foo = FooClient(client, ...)
bar = BarClient(foo, ...)
baz = BazClient(bar, ...)

pet = baz.pet.getPetById(petId=42).result()
```
Hear, the callbacks have no place in the client code, they should be internal to each ``{Foo|Bar|Baz}Client``.

If ``BarClient`` wants to instrument the response, it would do so in its ``BarOperationDecorator``.

```python
class BarOperationDecorator(object):
    def __init__(self, delegate):
        self.delegate = delegate
    ...
    def __call__(self, **op_kwargs):
        # Something like:
        # - add callback to request_options
        # - return delegate(**op_kwargs)
```

prat0318:
Nice point. I didn't think about it with respect to multiple decorator clients. My example would not let decorator client to work on actual `incoming_response` but on the value passed by the parent client in the chain.

analogue:
Ready to merge

CHANGELOG.rst
```diff
@@ -1,6 +1,14 @@
 Changelog
 =========
 
+7.0.0 (2015-XX-XX)
+------------------
+- Support per-request `response_callbacks <http://bravado.readthedocs.org/en/latest/configuration.html#per-request-configuration>`_
+  to enable ``SwaggerClient`` decorators to instrument an ``IncomingResponse``
+  post-receive. This is a non-backwards compatible change iff you have
+  implemented a custom ``HttpClient``. Consult the changes in signature to
+  ``HttpClient.request()`` and ``HttpFuture``'s constructor.
```
prat0318: nit: also add per request-wise `also_return_response` config addition

tests/http_future/HttpFuture/result_test.py
```diff
@@ -1,85 +1,84 @@
...
-        callback=response_callback)
+        operation=Mock(spec=Operation))
 
     assert 'hello world' == http_future.result()
 
 
-def test_non_2XX_with_response_callback():
-    response_adapter_instance = Mock(spec=IncomingResponse, status_code=400)
-    response_callback = Mock(side_effect=HTTPError(response_adapter_instance))
+@patch('bravado.http_future.unmarshal_response')
```
prat0318: `autospec=True` to the patch statements.

prat0318:
lg2m

CHANGELOG.rst
analogue: ack

tests/http_future/HttpFuture/result_test.py
analogue: ack

PR #184, opened 2015-11-19
Support for recursive refs and remote refs

analogue:
This basically pulls in the latest bravado-core and swagger-spec-validator so that Swagger specs with recursive refs work as expected.

Highlights:
- http_client now used by jsonschema's RefResolver to download remote refs
- Updated petstore spec to the latest and greatest

Testing Done:
- End to end test with internal happy hour service split into multiple json files

### DIFF
--- a/bravado/client.py
+++ b/bravado/client.py
@@ -175,16 +175,16 @@ def inject_headers_for_remote_refs(request_callable, request_headers):
     :param request_callable: method on http_client to make a http request
     :param request_headers: headers to inject when retrieving remote refs
     """
-    def request_wrapper(*args, **kwargs):
+    def request_wrapper(request_params, *args, **kwargs):
 
         def is_remote_ref_request(request_kwargs):
             # operation is only present for service calls
             return request_kwargs.get('operation') is None
 
         if is_remote_ref_request(kwargs):
-            request_params = args[0]
             request_params['headers'] = request_headers
-        return request_callable(*args, **kwargs)
+
+        return request_callable(request_params, *args, **kwargs)
 
     return request_wrapper
 
--- a/bravado/client.py
+++ b/bravado/client.py
@@ -53,6 +53,7 @@ from bravado_core.spec import Spec
 from six import iteritems, itervalues
 
 from bravado.docstring_property import docstring_property
+from bravado.http_client import HttpClient
 from bravado.requests_client import RequestsClient
 from bravado.swagger_model import Loader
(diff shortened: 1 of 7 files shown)


bravado/client.py
```diff
@@ -154,6 +168,27 @@ def __dir__(self):
...
+    :param request_headers: headers to inject when retrieving remote refs
+    """
+    def request_wrapper(*args, **kwargs):
+
+        def is_remote_ref_request(request_kwargs):
+            # operation is only present for service calls
+            return request_kwargs.get('operation') is None
+
+        if is_remote_ref_request(kwargs):
+            request_params = args[0]
```
prat0318: it will be better to add first param as `request_params` and rest as *args.

bravado/client.py
```diff
@@ -102,11 +102,23 @@ def from_url(cls, spec_url, http_client=None, request_headers=None,
...
         log.debug(u"Loading from %s" % spec_url)
         http_client = http_client or RequestsClient()
         loader = Loader(http_client, request_headers=request_headers)
         spec_dict = loader.load_spec(spec_url)
+
+        # RefResolver may have to download additional json files (remote refs)
+        # via http. Wrap http_client's request() so that request headers are
+        # passed along with the request transparently. Yeah, this is not ideal,
+        # but since RefResolver has new found responsibilities, it is
+        # functional.
```
prat0318: I have few doubts:
1) So, `spec_dict` at this point may have external refs present, correct?
2) What are the points when additional json files would need to be downloaded  (i suppose it is done in bravado-core and this http_client is fed to it)?
3) Instead of adding this wrapper, is it possible to use a clone of this `http_client` and use that instead in bravado-core?

analogue: 1) yes
2) bravado-core if spec validation is disabled; ssv if spec validation is enabled. Downloads only occur once because the resolver is shared between ssv and bravado-core.
3) What would a "clone of this http_client" look like in code? Why?

bravado/client.py
analogue: Fixed

bravado/client.py
prat0318: I was thinking more like unwrapping the wrapper once the work is finished (after the call to `bravado-core`, which guarantees that the work of wrapper is done), instead of having the check looking for `operation` key (which looks more hacky to me). I am not too sure on its implementation details but can think on that tomo.

PR #189, opened 2016-01-29
Support dot in operation name

cyprieng:
This pull request allow dot in an operation name. OperationID can have dot, especially if you use [connexion](https://github.com/zalando/connexion).

### DIFF
--- a/bravado/client.py
+++ b/bravado/client.py
@@ -195,23 +195,40 @@ class ResourceDecorator(object):
     operations can be instrumented.
     """
 
-    def __init__(self, resource):
+    def __init__(self, resource, base_name=''):
         """
         :type resource: :class:`bravado_core.resource.Resource`
         """
+        self.base_name = base_name  # Base name in case there is a dot in the operationId
         self.resource = resource
 
     def __getattr__(self, name):
         """
         :rtype: :class:`CallableOperation`
         """
-        return CallableOperation(getattr(self.resource, name))
+        try:
+            return CallableOperation(getattr(self.resource, '{0}{1}'.format(self.base_name, name)))
+        except AttributeError:  # Error in the attribute
+            if any(i.startswith(u'{0}.'.format(name)) for i in dir(self.resource)):
+                # There is an operation id that start with "name."
+                return ResourceDecorator(self.resource, '{0}{1}.'.format(self.base_name, name))
+            raise
 
     def __dir__(self):
         """
         Exposes correct attrs on resource when tab completing in a REPL
         """
-        return self.resource.__dir__()
+        dir_list = []
+
+        if not self.base_name == '':
+            # If there is a base_name, get all the element starting by base_name
+            for value in self.resource.__dir__():
+                if value.startswith(self.base_name):
+                    dir_list.append(value.replace(self.base_name, ''))
+        else:
+            return self.resource.__dir__()
+
+        return dir_list
 
 
 class CallableOperation(object):
--- a/bravado/client.py
+++ b/bravado/client.py
@@ -218,17 +218,11 @@ class ResourceDecorator(object):
         """
         Exposes correct attrs on resource when tab completing in a REPL
         """
-        dir_list = []
-
         if not self.base_name == '':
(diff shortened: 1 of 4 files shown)


analogue:
From the definition of an [operationId](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#operation-object), I'd say connexion is taking liberties and going a bit too far by embedding a period.

Quoting directly:

Tools and libraries MAY use the operationId to uniquely identify an operation, therefore, it is recommended to follow common programming naming conventions.

As such, which programming languages allows periods in their method names?

cyprien-g:
I am completely agree with you.
But I don't see the problem to support it, even if it not really compliant with the swagger specification.

analogue:
I'd much rather see the operation id sanitized instead of creating multiple levels of resources for no particular reason.

`my.poorly.named.operation.id` becomes `my_poorly_named_operation_id`.

This can easily be done in [bravado-core](https://github.com/Yelp/bravado-core/blob/5a17de000d1478c76dd548c6b4b6e0065520e0fc/bravado_core/operation.py#L77)

traxair:
Well even if it's unofficial and connexion is taking liberties, this is a very harmless fix, and they have more stars than you. We are contributing to your library that we use and like, and I consider such a rejection irresponsible in a young and growing community. Please reconsider :)

striglia:
Have we asked any core Swagger maintainers for clarification on this point? Seems like the spec is unhelpfully vague as to what's legal and not.

analogue:
@webron Opinions?

webron:
@analogue - I assume you mean opinions about the `operationId` and not some other comments that have been made here (which bother me even though I'm not a maintainer of this project).

The spec intentionally doesn't impose any limits. The `Tools and libraries MAY use the operationId to uniquely identify an operation, therefore, it is recommended to follow common programming naming conventions.` lines was added after some complaints from library maintainers. Fact is, it can be anything. It can be a string with space. The parsing library should be able to deal with it and not break. The added comment is to try and guide users to not do anything insane there (like using spaces), but it doesn't explicitly prohibit it.

Whether bravado should handle these edge cases or not is entirely up to the maintainers. I cannot say how easy or complicated it may be, not only to implement but also to maintain.

analogue:
@cyprien-g @traxair  Your proposal turns an operationId into one or more nested resources and then finally an operation. This is far from harmless. It breaks the interface for `ResourceDecorator` which returns `CallableOperation` for each operation.

I've given you a simple, sustainable, and backwards compatible solution in bravado-core without introducing unneeded complexity. What you decide to do with the suggestion is up to you :smiley:

cyprieng:
I think we should handle a period in the operationId. But I do not really care how this is done. The solution of @analogue can work, and we can handle other poorly named operationId in the same way.

The only problem with this solution is if there is two operationId like aa.bb and aa_bb, but it shouldn't happen.

So if everybody is agree with this solution, I can make a PR on bravado-core as @analogue suggest.

traxair:
If I may, we could do it the Pandas way. In Pandas, column names are arbitrary, but you can access them using `__getattr__` or `__getitem__`. The latter actually takes care of corner cases. You could consider that `recourcedecorator_instance.operation.with.a.dot` in unacceptable and won't work (and I agree with you), however `recourcedecorator_instance['operation.with.a.dot']` would provide a safe you to go through corner cases, including the case of operationId with spaces (though @webron said it was insane, but kind of not forbidden)

cyprieng:
@analogue So which way do you prefer ? Sanitize operationId or use @traxair method ?

analogue:
@cyprieng sanitize works for me

traxair:
For the sake of argumentation and keeping the behavior of bravado
predictable, why is the other solution not suiting you?

webron:
Sanitization has the advantage that it can make tools that consume the spec more robust. While it is the tools' responsibility to deal with these characters in the `operationId`, it just makes their lives easier (or rather their developers' lives easier).

cyprieng:
Ok, fine for me. I will make a PR soon.

PR #192, opened 2016-02-29
Fix datetime when passed as a param

### LINKED ISSUE #191: Passing datetime as request body throws JSON serializable error.
Relates to #80. I am not sure it got resolved.

The flow here deals only for swagger_to_py flow and not the other way round. If there is a date-time param in request body, it is not currently transformed to a str to make `json.dumps` work.

`SwaggerTypeCheck` currently is being used both for to_wire and from_wire flow. This needs to be distinguished.

prat0318:
* docstrings and tests are yet to be added.

It checks if it is a `to_wire` flow, in that case, converts `datetime` to `str`.

Fixes #191


swaggerpy/__init__.py
```diff
@@ -1,2 +1,2 @@
 
-version = '0.7.10'
```
analogue: What happened to 0.7.11?

prat0318: oh i think 0.7.11 was not released. will merge 12 with 11 changes.

analogue:
:+1:

prat0318:
Ack. Waiting for travis to be green.

PR #196, opened 2016-03-03
Deleting logic for handling form data in favor of python 'requests' for the Asynchronous client

### LINKED ISSUE #194: TypeError with form multi-part data when using the AsynchronousHttpClient
I am getting the following error when I try sending multi-part form data (such as a file and some strings/integers):
`
File "/code/virtualenv_run/local/lib/python2.7/site-packages/swaggerpy/response.py", line 55, in __init__
    self._request = self._http_client.start_request(request_params)
  File "/code/virtualenv_run/local/lib/python2.7/site-packages/swaggerpy/async_http_client.py", line 50, in start_request
    'bodyProducer': stringify_body(request_params),
  File "/code/virtualenv_run/local/lib/python2.7/site-packages/swaggerpy/async_http_client.py", line 159, in stringify_body
    data = create_multipart_content(request_params, headers)
  File "/code/virtualenv_run/local/lib/python2.7/site-packages/swaggerpy/multipart_response.py", line 47, in create_multipart_content
    return "\r\n".join(lines)
TypeError: sequence item 7: expected string or Unicode, int found
`

I believe it is due to try joining integers and strings together ([see this line] (https://github.com/Yelp/bravado/blob/swaggerpy/swaggerpy/multipart_response.py#L21)).
A fix could be:
`lines.extend(["--" + boundary, header, "", str(content)])`

working on a patch right now

lucagiovagnoli:
fixes #194 and adds some tests for the multipart_response.py file


lucagiovagnoli:
fixes both #197 and #194 bugs by replacing the logic responsible for creating the request with the one in python package "request".

swaggerpy/async_http_client.py
```diff
@@ -21,12 +22,9 @@
 from twisted.web.client import Agent
 from twisted.web.client import FileBodyProducer
 from twisted.web.http_headers import Headers
-from yelp_uri import urllib_utf8
```
laucia: you probably can remove the need for "yelp_uri >= 1.1.0" in setup.py

tests/async_http_client_test.py
```diff
@@ -115,7 +45,7 @@ def test_success_AsyncHTTP_response(self):
         ) as mock_Async:
             req = {
                 'method': 'GET',
-                'url': 'foo',
+                'url': 'http://foo',
```
laucia: maybe we should be trying to add this automatically

swaggerpy/async_http_client.py
laucia: and remove the dependency used just for debug in client.py as well

lucagiovagnoli: fixed but had to port some helpers inside. still better than preventing people from using this because of yelp_uri.

tests/async_http_client_test.py
lucagiovagnoli: I think I'd prefer leaving it as is, as it is the convention used by python 'requests'.

swaggerpy/client.py
```diff
@@ -561,3 +559,37 @@ def stringify_body(value):
...
+    return x.iteritems() if isinstance(x, dict) else x
+
+
+def _utf8_urlencode(query, *args, **kwargs):
+    """A wrapper for urllib.urlencode() that UTF-8 encodes query if
+    query is unicode. Always returns a str."""
+    return urllib.urlencode(
+        [(_to_utf8(k), _to_utf8(v)) for (k, v) in _pairs(query)],
+        *args, **kwargs
+    )
```
prat0318: can we move these function to a separate module

prat0318:
:+1:

swaggerpy/client.py
lucagiovagnoli: Actually I just found out that yelp_uri is open source https://github.com/Yelp/yelp_uri so I am putting back the dependency, anybody can use it and we don't duplicate code :)

prat0318:
lgtm. Can you bump the version and add an entry to changelog. should be good to merge then

prat0318:
Merged.

Issue #168, opened 2015-08-11
Instantiate a client from a .yml definition instead of .json only


stratosgear:
It should be easy no?

analogue:
Pull requests with unit tests happily accepted :-)

PR #198, opened 2016-03-06
Add support for parsing YAML specs

sjaensch:
My goal is to stay compatible with the YAML specs in the [OpenAPI repository](https://github.com/OAI/OpenAPI-Specification/tree/master/examples/v2.0/yaml). This necessitates the cast of HTML status codes to strings, since otherwise there will be an exception when trying to use the parsed spec.

Closes Yelp/bravado#168

### DIFF
--- a/bravado/swagger_model.py
+++ b/bravado/swagger_model.py
@@ -119,7 +119,10 @@ class Loader(object):
             for method, operation in iter(methods.items()):
                 if 'responses' in operation:
                     operation['responses'] = dict(
-                        (str(code), response) for code, response in iter(operation['responses'].items())
+                        (str(code), response)
+                        for code, response in iter(
+                            operation['responses'].items()
+                        )
                     )
 
         return data
--- a/bravado/swagger_model.py
+++ b/bravado/swagger_model.py
@@ -2,7 +2,10 @@
 import contextlib
 import logging
 import os
+import os.path
+import yaml
 
+from bravado_core.spec import is_yaml
 from six.moves import urllib
 from six.moves.urllib import parse as urlparse
 
@@ -24,22 +27,25 @@ class FileEventual(object):
     class FileResponse(object):
 
         def __init__(self, data):
-            self.data = data
+            self.text = data
+            self.headers = {}
 
         def json(self):
-            return self.data
+            return self.text
 
     def __init__(self, path):
         self.path = path
+        self.is_yaml = is_yaml(path)
 
     def get_path(self):
-        if not self.path.endswith('.json'):
+        if not self.path.endswith('.json') and not self.is_yaml:
             return self.path + '.json'
         return self.path
 
     def wait(self, timeout=None):
         with contextlib.closing(urllib.request.urlopen(self.get_path())) as fp:
-            return self.FileResponse(json.load(fp))
+            content = fp.read() if self.is_yaml else json.load(fp)
+            return self.FileResponse(content)
 
     def result(self, *args, **kwargs):
         return self.wait(*args, **kwargs)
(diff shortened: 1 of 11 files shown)


bravado/swagger_model.py
```diff
@@ -86,12 +92,40 @@ def load_spec(self, spec_url, base_url=None):
...
-        return spec_json
+        ).result()
+
+        content_type = response.headers.get('content-type', '').lower()
+        if is_yaml(spec_url, content_type):
+            return self.load_yaml(response.text)
+        else:
+            return response.json()
+
+    def load_yaml(self, text):
```
bpicolo: Can we just load them all as if they're yaml? Json is a subset

bravado/swagger_model.py
```diff
@@ -86,12 +92,40 @@ def load_spec(self, spec_url, base_url=None):
...
+        """Load a YAML Swagger spec from the given string, transforming
```
bpicolo: This suggests to me that your YAML spec is incorrect, we shouldn't need to cast these to string. Need to wrap in quotes.
https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#responses-object-example

bpicolo: Ahh, gotcha, sounds like specs out there aren't well formed in general and yaml might be inconsistent. Cool

bravado/swagger_model.py
sjaensch: Yes but YAML is not a JSON subset, we'd be potentially parsing/ingesting files that are not valid JSON.

bpicolo: Hmmm I guess that's fair.

bravado/swagger_model.py
```diff
@@ -24,22 +27,25 @@ class FileEventual(object):
...
     def get_path(self):
-        if not self.path.endswith('.json'):
+        if not self.path.endswith('.json') and not self.is_yaml:
             return self.path + '.json'
         return self.path
 
     def wait(self, timeout=None):
         with contextlib.closing(urllib.request.urlopen(self.get_path())) as fp:
-            return self.FileResponse(json.load(fp))
+            content = fp.read() if self.is_yaml else json.load(fp)
```
analogue: For the yaml case, why is this setting `content` to the raw contents instead of `yaml.load(fp)`?  I actually think `content` would benefit from being renamed since in both cases, it is the Python object that represents the ingested json or yaml file.

tests/swagger_model/load_file_test.py
```diff
@@ -8,7 +8,26 @@ def test_success():
...
+    ),
+)
+def test_success_yaml(filename):
+    spec_yaml = load_file(filename)
+    assert '2.0' == spec_yaml['swagger']
+
+
+def test_spec_internal_representation_identical():
+    spec_json = load_file('test-data/2.0/petstore/swagger.json')
+    spec_yaml = load_file('test-data/2.0/petstore/swagger.json')
```
analogue: s/swagger.json/swagger.yaml

bravado/swagger_model.py
sjaensch: It's because it's mimicking a response object, which has a json() method to return a Python representation of JSON content. So in Loader.load_spec we distinguish between the two cases and either load YAML ourselves or just call response.json().

The FileResponse object was originally built for the JSON case, so it simply parsed JSON immediately and then returned the parsed data when json() is called. I'm going to get rid of this and separate out the two steps of reading JSON content from disk and then parsing it.

analogue:
:+1:

tests/swagger_model/load_file_test.py
sjaensch: Thanks for catching this by the way. ;)

PR #200, opened 2016-03-18
Use yaml.safe_load instead of yaml.load

sjaensch:


lucagiovagnoli:
Looks good to me - merging

PR #218, opened 2016-04-28
Fido 3.0.0 - EventualResult instead of concurrent futures - Fix requests unicode bug

lucagiovagnoli:
**Upgrade bravado to use fido 3.0.0**
This change makes bravado capable of handling EventualResult objects which are now returned by Fido instead of a concurrent.futures.
I acted on our FutureAdaptor layer of abstraction by creating FidoFutureAdapter and making it and RequestFutureAdapter 'implement' FutureAdapter.

**Requests bug**
I also introduced a minor little workaround caused by a bug in 'requests' version < 2.8.1  which is documented in their changelog at https://pypi.python.org/pypi/requests

**Tests**
Testing might be broken on github until fido gets updated to 3.0.0 in https://pypi.python.org/simple/fido/
I manually tested with fido 3.0.0 and our internal acceptance suite though and it was all green.

### DIFF
--- a/bravado/http_future.py
+++ b/bravado/http_future.py
@@ -8,12 +8,33 @@ import six
 from bravado.exception import HTTPError
 
 
-class HttpFuture(object):
-    """Wrapper for a :class:`concurrent.futures.Future` that returns an HTTP
-    response.
+class FutureAdapter(object):
+    """
+    Mimics a :class:`concurrent.futures.Future` regardless of which client is
+    performing the request, whether it is synchronous or actually asynchornous.
 
-    :param concurrent_future: The concurrent concurrent_future to wrap.
-    :type concurrent_future: :class: `concurrent.futures.Future`
+    This adapter must be implemented by all bravado clients such as FidoClient
+    or RequestsClient to wrap the object returned by their 'request' method.
+
+    """
+
+    def result(self, timeout=None):
+        """
+        Must implement a result method which blocks on result retrieval.
+
+        :param timeout: maximum time to wait on result retrieval. Defaults to
+            None which means blocking undefinitely.
+        """
+        raise NotImplementedError(
+            "FutureAdapter must implement 'result' method"
+        )
+
+
+class HttpFuture(object):
+    """Wrapper for a :class:`FutureAdapter` that returns an HTTP response.
+
+    :param future: The future object to wrap.
+    :type future: :class: `FutureAdapter`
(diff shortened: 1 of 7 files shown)


bravado/http_future.py
```diff
@@ -8,12 +8,33 @@
 from bravado.exception import HTTPError
 
 
+class FutureAdapter(object):
+    """
+    Mimics a :class:`concurrent.futures.Future` regardless of which client is
+    performing the request, whether it is synchronous or actually asynchornous.
+
+    This adapter must be implemented by all bravado clients such as FidoClient
+    or RequestsClient to wrap the object returned by their 'request' method.
```
laucia: I don't see request clients using it though

lucagiovagnoli: that's just the interface,
RequestsClient is using it [here](https://github.com/Yelp/bravado/pull/218/files#diff-0c505f6596e107bab9531f24ef030343R198)
```python
class RequestsFutureAdapter(FutureAdapter):
```

 and FidoClient is using it [here](https://github.com/Yelp/bravado/pull/218/files#diff-4e6e269b35d40d26025c53fbc082cc28R134)
```python
class FidoFutureAdapter(FutureAdapter):
```
:)

lucagiovagnoli:
I just uploaded fido==3.0.0 to https://pypi.python.org/pypi/fido  
tests all green

askedrelic:
lg2m

bravado/http_future.py
```diff
@@ -8,12 +8,33 @@
...
+
+        :param timeout: maximum time to wait on result retrieval. Defaults to
+            None which means blocking undefinitely.
+        """
+        raise NotImplementedError(
+            "FutureAdapter must implement 'result' method"
+        )
+
+
 class HttpFuture(object):
```
laucia: shouldn't this inherit from FutureAdapter?

bravado/requests_client.py
```diff
@@ -137,7 +138,8 @@ def request(self, request_params, operation=None, response_callbacks=None,
         requests_future = RequestsFutureAdapter(
             self.session,
             self.authenticated_request(sanitized_params),
-            misc_options)
+            misc_options
```
laucia: trailing commas ?

laucia:
lg2m, just a formating nitpick

bravado/requests_client.py
lucagiovagnoli: keeping the style - amended

PR #224, opened 2016-05-20
Pinpoint fido version

lucagiovagnoli:
Avoid breaking the build in case of backward incompatible changes (fido 3.2.0 should be py3 ready)


laucia:
lg2m

sjaensch:
lg2m2

PR #226, opened 2016-05-24
fix required fields for header

necrolyte2:
I'm sure this is not the correct way to fix this, but I was not sure

When you have this in one of your swagger parameters
```
{                                                                                                                                                                                                                                                                       
            "name": "Authorization",                                                                                                                                                                                                                                              
            "in": "header",                                                                                                                                                                                                                                                       
            "description": "Basic Auth access token.",                                                                                                                                                                                                                            
            "required": true,                                                                                                                                                                                                                                                     
            "type": "string"                                                                                                                                                                                                                                                      
},    
```

you will get 

```
SwaggerMappingError: Authorization is a required parameter
```

even with

```
headers = {                                                                                                                                                                                                                                                                                                                                                                                                                                                                                          
            'Authorization': 'asdfaf'                                                                                                                                                                                                                                   
}
create_response = self.client.testthing.post_testthing(                                                                                                                                                                                                   
            body=self.thing,
            _request_options={                                                                                                                                                                                                                                                    
                "headers": headers                                                                                                                                                                                              
            }                                                                                                                                                                                                                                                                     
)             
```

### DIFF
--- a/bravado/client.py
+++ b/bravado/client.py
@@ -314,7 +314,7 @@ def construct_params(operation, request, op_kwargs):
 
     # Check required params and non-required params with a 'default' value
     for remaining_param in itervalues(current_params):
-        if remaining_param.required:
+        if remaining_param.required and remaining_param.location != 'header':
             raise SwaggerMappingError(
                 '{0} is a required parameter'.format(remaining_param.name))
         if not remaining_param.required and remaining_param.has_default():


macisamuele:
Many thanks for your contribution.
I tried to figure out what was your original problem, but I'm not able to reproduce it.
Could you check if the problem still affects you?
If so could you give me more context in how to reproduce it, like the whole swagger specs and the test python file that you have used? Thanks a lot

necrolyte2:
See if this Gist works for you
I had to extract only the pieces as I could not post the entire swagger and code so hopefully I have extracted everything that is needed
https://gist.github.com/necrolyte2/e4f8f4f76dd3c5637af8e6cf77a73ca5

This is the error I get when I run it with the latest bravado client(8.3.0)
```
  File "bravado_test.py", line 157, in <module>
    create_container(client, headers, **body)
  File "bravado_test.py", line 131, in create_container
    "response_callbacks": [print_response, log_response]
  File "/Users/user/miniconda2/lib/python2.7/site-packages/bravado/client.py", line 250, in __call__
    self.operation, request_options, **op_kwargs)
  File "/Users/user/miniconda2/lib/python2.7/site-packages/bravado/client.py", line 291, in construct_request
    construct_params(operation, request, op_kwargs)
  File "/Users/user/miniconda2/lib/python2.7/site-packages/bravado/client.py", line 319, in construct_params
    '{0} is a required parameter'.format(remaining_param.name))
bravado_core.exception.SwaggerMappingError: accept is a required parameter
```

macisamuele:
@necrolyte2 thanks for having provided the Gist ... it helped me a lot to understand what was your point.

The exception that you're getting is related to the fact that you have not correctly specified the parameters.

On [line 130](https://gist.github.com/necrolyte2/e4f8f4f76dd3c5637af8e6cf77a73ca5#file-test-py-L130) you're trying to inject the headers in the requests, but those will be added directly on the request object which is not checked against the specs, so you get ``SwaggerMappingError``.

### Workaround
If you try instead to insert the header parameters like you did for ``body`` and ``key`` you will successfully run your request.

```python
create_response = client.create.create(
    body=create,
    key='apikey',
    _request_options={
        # "headers": headers,  # look two line below :)
        "response_callbacks": [print_response, log_response]
    },
    **headers  # dictionary expansion for injecting the header from the dictionary
)
```

### More general solution (what I prefer more)
Your code modification will achieve the goal of removing the annoying exception but basically not validate the presence of the required headers parameters.

I would do prefer something like (in case the parameter is a required one):
 - check if the parameter is an header parameter
    - if so check that is present and compliant with the specs
    - if not raise ``SwaggerMappingError```

So you'll enforce also the correct validation of the values injected in the request through ``_request_options``


### Suggestion (a little off-topic)
Checking your specs I have seen that you're using ``security`` specs but your specs are not defining any ``securityDefinitions`` and it is not compliant with the [Swagger Specs](swagger.io/specification/#securityRequirementObject).

necrolyte2:
@macisamuele Thanks!

Your Workaround solution seems like the correct way for us to go. There is no reason that we would need to put them into the `_request_options`
I think it would jus be easier to close this PR since the use case is really not really correct now I know how to correctly pass header arguments.

I thought I looked through the docs and didn't find where that was, but it has been a few months.

Thanks for the suggestion as well for the spec.

Issue #223, opened 2016-05-19
Python3 support for FidoClient


struys:
The FidoClient raises an exception if imported under python3:
```python
if six.PY3:
    raise ImportError("The fido client is not yet supported in py3")
```

I can't find documentation explaining why it doesn't support python3 so I figured i'd file this to get answer and start enumerating the blockers.

cc/@lucagiovagnoli @jnb

lucagiovagnoli:
Mmh, I think that's a wider issue about Twisted not being python 3 ready. See this comment [on bravado](https://github.com/Yelp/bravado/blob/master/tests/fido_client/FidoClient/prepare_request_for_twisted_test.py#L20):
```python
@pytest.mark.skipif(
    six.PY3,
    reason="Fido client is not usable in py3 until twisted supports it",
)
```

Now that [Twisted is py3 ready](https://pypi.python.org/pypi/Twisted), it should not be hard to make fix this :)

lucagiovagnoli:
https://github.com/Yelp/fido/pull/34

PR #228, opened 2016-06-01
Bravado can use fido python 3 ready

lucagiovagnoli:
Enabling use of the new Fido client python3 ready. Let's have bravado use it.

I upgraded bottle to a more recent version where [issue221](https://github.com/gabrielfalcao/HTTPretty/issues/221) is fixed.
Using now bytes for the request body as requested by fido 3.2.0 

@akanak

### DIFF
--- a/bravado/fido_client.py
+++ b/bravado/fido_client.py
@@ -2,7 +2,6 @@
 import logging
 
 import requests
-import six
 
 import fido
 from bravado_core.response import IncomingResponse
@@ -117,7 +116,8 @@ class FidoClient(HttpClient):
             # using requests < 2.8.1 due to a bug while handling unicode values
             # See changelog 2.8.1 at https://pypi.python.org/pypi/requests
             'method': str(prepared_request.method or 'GET'),
-            'body': to_bytes(prepared_request.body) if prepared_request.body is not None else None,
+            'body': to_bytes(prepared_request.body) if prepared_request.body \
+            is not None else None,
             'headers': prepared_request.headers,
             'url': prepared_request.url,
         }
--- a/bravado/fido_client.py
+++ b/bravado/fido_client.py
@@ -116,8 +116,10 @@ class FidoClient(HttpClient):
             # using requests < 2.8.1 due to a bug while handling unicode values
             # See changelog 2.8.1 at https://pypi.python.org/pypi/requests
             'method': str(prepared_request.method or 'GET'),
-            'body': to_bytes(prepared_request.body) if prepared_request.body \
-            is not None else None,
+            'body': (
+                to_bytes(prepared_request.body)
+                if prepared_request.body is not None else None
+            ),
             'headers': prepared_request.headers,
             'url': prepared_request.url,
         }
--- a/bravado/fido_client.py
+++ b/bravado/fido_client.py
@@ -9,6 +9,7 @@ from bravado_core.response import IncomingResponse
 from bravado.http_client import HttpClient
 from bravado.http_future import FutureAdapter
 from bravado.http_future import HttpFuture
+from yelp_bytes import to_bytes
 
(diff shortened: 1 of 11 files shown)


tests/fido_client/FidoClient/prepare_request_for_twisted_test.py
```diff
@@ -4,22 +4,24 @@
...
+from bravado.fido_client import FidoClient
 
-try:
-    from bravado.fido_client import FidoClient
-except ImportError:
-    FidoClient = Mock()  # Tests will be skipped in py3
+
+def test_prepare_request_for_twisted_get():
+    request_params = dict(
+        url='http://example.com/api-docs')
```
laucia:
```
{
    'url': 'http://example.com/api-docs'
}
```

bravado/fido_client.py
```diff
@@ -119,7 +116,8 @@ def prepare_request_for_twisted(request_params):
             # using requests < 2.8.1 due to a bug while handling unicode values
             # See changelog 2.8.1 at https://pypi.python.org/pypi/requests
             'method': str(prepared_request.method or 'GET'),
-            'body': prepared_request.body,
+            'body': to_bytes(prepared_request.body) if prepared_request.body \
```
laucia: I am not a huge fan of the line return here, can you rewrite this with parenthesis rather ?
```
(
    to_bytes(prepared_request.body) if prepared_request.body
    if prepared_request.body is not None
    else None
)
```
That's really accessory though

akanak:
lg2m.

bravado/compat.py
```diff
@@ -1,5 +0,0 @@
-
-try:
```
laucia: Why remove this? simplejson is more performant than the stdlib json.

laucia:
lg2m

setup.py
```diff
@@ -32,7 +32,8 @@
     install_requires=[
         "bravado-core >= 4.2.2",
         "crochet >= 1.4.0",
-        "fido >= 2.1.0, < 3.2.0",
+        "fido >= 2.1.0, <= 3.2.0",
```
sjaensch: This is kind of a weird restriction. Do we know that bravado is incompatible with fido 3.2.1? Otherwise I would specify this with "< 3.3" and expect fido not to break compatibility in minor / patch releases. We could also simply remove this restriction and only add it back in if a known incompatibility comes up. We're not restricting other libraries in this way either.

lucagiovagnoli: I originally added it when I was porting fido to python 3 because I knew it would break it. I'm fine with < 3.3 but I would keep the restriction as fido is rapidly growing and in a highly dynamic state

sjaensch: Well, are we expecting further major breakages after the rewrite? It's these (premature?) restrictions that cause upgrade troubles for larger projects. It's why we're stuck on an old requests version even though the newest one seems to work just fine. If you use tools that do proper restriction checking (like pip-faster does) then you won't be able to use a newer version. pip doesn't really do sophisticated checking.

I'd vote for removing this restriction if there is no reason for it right now and instead focus on providing a stable API with fido. That's what we should aim for, that's what we should expect. Let's not code for failure. And if we do need to add such a restriction it should be for a major version, i.e. fido < 4.

lucagiovagnoli: I'm expecting another change of contract when implementing a timeout mechanism (planning on providing a timeout at client creation to be used as a default for http calls).
But I see your point. This restriction is not necessary but just a safety net, removing.

sjaensch:
lg2m otherwise, nice work! 👍

lucagiovagnoli:
This fixes #223

asottile:
seems you didn't actually release properly, 8.3.0 seems to not contain this change

PR #232, opened 2016-06-22
Pin cryptography to 1.3.4 so we can build on lucid

dannyperson:
Our Jenkins tests bravado on lucid, and has been failing to install the latest cryptography because the latest release drops support for OpenSSL 0.9.8, the newest release for lucid. This pins cryptography 1.3.4.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -4,3 +4,4 @@ httpretty==0.8.14
 mock<1.1.0
 pre-commit
 pytest
+cryptography==1.3.4


sjaensch:
This is a tricky one. First of all requirements-dev.txt only contains additional packages needed for developing bravado (like packages required for running the tests). cryptography is not one of them. This highlights a problem of Python packaging: we can't specify dependencies based on Linux distribution version. The same issue arises if you try to use bravado on Python 2.6: it will fetch Twisted which will error out on installation since it doesn't support Python 2.6 anymore.

I think this is something you need to fix locally. What we do internally is run our own PyPi server that only contains the package versions we want to use. Otherwise it'd be way too restrictive for users that do put in the work to stay up to date with their software stack. We don't want them to be stuck on an old cryptography version (or on an old Twisted version).

sjaensch:
Closing this PR as we've resolved the issue internally.

PR #233, opened 2016-06-23
Relax dev requirements; we should try not to pin these

sjaensch:
Tests pass locally with mock 2.0. I also don't think we have a hard dependency on the latest version of httpretty.


lucagiovagnoli:
_
    /|\
   /_|_\
 ____|____
 \_o_o_o_/
~~ |     ~~~~~
___t_________

PR #236, opened 2016-07-25
FidoClient backward-compatible with Swaggerpy

lucagiovagnoli:
We know that AsynchronousHttpClient in swaggerpy is unstable (it was even mentioned by @sjaensch and @laucia at europython talks), so there is not really a way to go async with using swaggerpy. *The suggested way to proceed is to upgrade to bravado and swagger2.0.*

As this could take some time, in the meantime I am proposing a change to make FidoClient backward-compatible with swaggerpy. I am posting the code change without tests to collect some feedback or discuss about it.

### DIFF
--- a/bravado/fido_client.py
+++ b/bravado/fido_client.py
@@ -5,9 +5,11 @@ import requests
 
 import fido
 from bravado_core.response import IncomingResponse
+from bravado.exception import HTTPError
 from bravado.http_client import HttpClient
 from bravado.http_future import FutureAdapter
 from bravado.http_future import HttpFuture
+from bravado.http_future import SwaggerpyLegacyHttpFuture
 from yelp_bytes import to_bytes
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


sjaensch:
This pull request should be against the swaggerpy branch, not master, right?

lucagiovagnoli:
mmh no, because this is making FidoClient backward compatible with swaggerpy (adding the method start_request). Another option could be to make swaggerpy able to use FidoClient which I estimate could result in a much bigger refactoring

sjaensch:
@lucagiovagnoli explained it to me in-person. swaggerpy can, by adding a dependency to bravado, use the FidoClient with these changes. Moving the FidoClient into its own repository or adding a copy to swaggerpy would mean a much bigger pull request with quite a bit of duplicated code. So I'm fine with merging this since swaggerpy users gain so much from it. 🛳

tomelm:
I'm supportive of this, would be nice to have non-blocking async calls as part of our legacy support for swagger12.

lucagiovagnoli:
I ran this code change through our thorough acceptance test suite. Unfortunately I discovered quite a few issues.
1. It looks like I am getting a preposterous amount of Timeout errors.
2. swaggerpy error structures are very different from bravado ones so more effort must be spent on making those compatible.

I don't think the change is worth the effort. This was only born as a quick hack for swaggerpy to gain access to the asynchronous feature but it appears more tricky than expected. Plus I'd prefer not "polluting" bravado so much.

PR #238, opened 2016-08-09
correctly handle async POST requests in swaggerpy + py3

mxmul:
`AsynchronousHTTPClient.start_request` fails with a TypeError when the request has any form data in python3, because we assume that the body of the request is a byte string:

```
self = <class 'swaggerpy.async_http_client.AsynchronousHttpClient'>(), request_params = {'data': {'name': 'Matt'}, 'headers': {}, 'method': b'POST', 'url': 'http://localhost:49111/post'}

    def start_request(self, request_params):
        """Sets up the request params as per Twisted Agent needs.
            Sets up crochet and triggers the API request in background

            :param request_params: request parameters for API call
            :type request_params: dict

            :return: crochet EventualResult
            """

        prepared_request = requests.PreparedRequest()
        prepared_request.prepare(
            headers=request_params.get('headers'),
            data=request_params.get('data'),
            params=request_params.get('params'),
            files=request_params.get('files'),
            url=request_params.get('url'),
            method=request_params.get('method')
        )

        request_for_crochet = {
            'method': prepared_request.method or 'GET',
            'bodyProducer': FileBodyProducer(BytesIO(prepared_request.body))
>           if prepared_request.body else None,
            'headers': listify_headers(prepared_request.headers),
            'uri': prepared_request.url,
        }
E       TypeError: 'str' does not support the buffer interface
```

This branch encodes the request body to a byte string if it's unicode.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,5 +1,8 @@
 [report]
 exclude_lines =
+    # Have to re-enable the standard pragma
+    \#\s*pragma: no cover
+
     def __repr__
     raise AssertionError
     raise NotImplementedError
--- a/swaggerpy/async_http_client.py
+++ b/swaggerpy/async_http_client.py
@@ -55,10 +55,18 @@ class AsynchronousHttpClient(http_client.HttpClient):
             method=request_params.get('method')
         )
 
+        if isinstance(
+            prepared_request.body,
+            six.text_type,
+        ):  # pragma: no cover (PY2)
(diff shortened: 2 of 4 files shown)


prat0318:
👍 lgtm

lucagiovagnoli:
If possible, I would suggest not using the AsynchronousHttpClient which comes with swaggerpy because I found it to be unstable on production. Also, swaggerpy in general as it is not actively maintained as bravado (and handles old swagger 1.x specs).

PR #247, opened 2016-08-29
Fix build breakage caused by incorrect specs related to security schema

sjaensch:
bravado-core added partial support for security schemas recently. We shouldn't modify the petstore spec. I could change the tests to use a different endpoint, but I think incorporating some of the security stuff into the tests is not that bad.


sjaensch:
Going to merge this in to unblock development.

PR #249, opened 2016-09-01
Fix for issue 222

helgridly:
Raise `HTTPError` without attempting to unmarshal the response if the status code isn't 2xx.

### DIFF
--- a/bravado/http_future.py
+++ b/bravado/http_future.py
@@ -70,18 +70,17 @@ class HttpFuture(object):
         inner_response = self.future.result(timeout=timeout)
         incoming_response = self.response_adapter(inner_response)
 
-        if self.operation is not None:
-            unmarshal_response(
-                incoming_response,
-                self.operation,
-                self.response_callbacks)
-
-            swagger_result = incoming_response.swagger_result
-            if self.also_return_response:
-                return swagger_result, incoming_response
-            return swagger_result
-
         if 200 <= incoming_response.status_code < 300:
+            if self.operation is not None:
+                unmarshal_response(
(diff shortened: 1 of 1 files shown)


sjaensch:
Thank you for the contribution! I think this is going too far though. We have services in production that return properly formatted JSON responses for e.g. 403, 404 HTTP status codes. We should either limit this to >= 500 or maybe check if there is a response defined in the spec for the status code, and if we do then unmarshal.

macisamuele:
I would prefer to not run the ``unmarshal_response`` method only if the operation object has no knowledge about the response format or if the content type is not application/json (#250).

The swagger specs allow you to define the response format also for not HTTP/200 cases, like in the [Operation Object Example](http://swagger.io/specification/#operationObject) (where is defined the format of the HTTP/405 case).

I believe that checking that ``incoming_response.content_type`` is a JSON could be enough.
```python
if self.operation is not None and incoming_response.content_type is 'application/json':
```

helgridly:
The "throw an `HTTPError` if there's no response format defined" idea sounds like a good one. My only concern is that users (i.e. me) now have to handle non-2xx codes differently, based on whether the server returns JSON or not - and that one case looks "normal" while the other raises an exception. Does this sound like a problem to you? Should we just return the incoming response for unmarshalable non-2xx codes too, and skip the exception entirely?

helgridly:
I dug into what's happening in my case some more and discovered I've run into a discrepancy between what Swagger describes and what we actually return. I'm happy for this to be closed if it's unhelpful, since it's user error.

PR #253, opened 2016-09-07
Enforced header validation (set through _request_options)

macisamuele:
I got inspiration by the problem shown by @necrolyte2 on #226 .
Trying to understand his issue I figured out that bravado is not validating the headers injected by ``_request_options``.

That's generally fine, since bravado validates that what is described on the swagger specs is present on the request and response. But @necrolyte2 was injecting some header parameters through ``_request_options`` and bravado was throwing him an exception (``SwaggerMappingError``) because some required header parameters were not set in the parameter list.

The goal of the PR is to add checks on the header parameters, specified in swagger specs, defined through ``_request_options`` instead of [``op_kwargs``](https://github.com/Yelp/bravado/blob/master/bravado/client.py#L295).

The code change should be considered as a backcompatible change.

### DIFF
--- a/tests/client/construct_params_test.py
+++ b/tests/client/construct_params_test.py
@@ -26,7 +26,10 @@ def test_no_params(mock_marshal_param, minimal_swagger_spec, request_dict):
         minimal_swagger_spec, '/pet/{petId}', 'get', {}))
     construct_params(op, request_dict, op_kwargs={})
     assert 0 == mock_marshal_param.call_count
-    assert 0 == len(request_dict)
+    assert request_dict == {
+        'params': {},
+        'headers': {},
+    }
 
 
 def test_extra_parameter_error(minimal_swagger_spec, request_dict):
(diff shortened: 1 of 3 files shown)


tests/client/construct_params_test.py
```diff
@@ -47,6 +50,16 @@ def test_required_parameter_missing(
...
 
 
+def test_validate_header_parameter_from_request_options(
+        minimal_swagger_spec, getPetById_spec, request_dict):
+    request_dict['url'] = '/pet/{petId}'
+    request_dict['headers']['api_key'] = 'api_key'
+
+    op = CallableOperation(Operation.from_spec(
+        minimal_swagger_spec, '/pet/{petId}', 'delete', getPetById_spec))
+    construct_params(op, request_dict, op_kwargs={'petId': 1})
```
sjaensch: Could you please mock marshal_param and add an assertion about how many times it has been called, just like in the test below?

PR #254, opened 2016-09-09
Add more granular exception classes for HTTPError

### LINKED ISSUE #252: Better HTTP Exceptions?
It would be nice to have a little more granular HTTP Exceptions, rather than just a single one.

All HTTPExceptions could inherit from the existing HTTPException class, but we could have ones like:

NotFoundHTTPException(HTTPException):
ForbiddenHTTPException(HTTPException):
GatewayTimeoutHTTPException(HTTPException):

It would be nice for when I'm catching different exceptions and trying to handle each one differently.

bxm156:
Add some common HTTP exception classes that inherit from HTTPError.
Added a common 4xx and 5xx error classes to start with.
fixes #252

### DIFF
--- a/bravado/exception.py
+++ b/bravado/exception.py
@@ -1,7 +1,24 @@
 # -*- coding: utf-8 -*-
+from six import with_metaclass
 
 
-class HTTPError(IOError):
+status_map = {}
+
+
+def _register_exception(exception_class):
+    status_map[exception_class.status_code] = exception_class
+
+
+class HTTPErrorType(type):
+
+    def __new__(cls, *args, **kwargs):
+        new_class = super(HTTPErrorType, cls).__new__(cls, *args, **kwargs)
+        if hasattr(new_class, 'status_code'):
+            _register_exception(new_class)
+        return new_class
+
+
+class HTTPError(with_metaclass(HTTPErrorType, IOError)):
     """Unified HTTPError used across all http_client implementations.
     """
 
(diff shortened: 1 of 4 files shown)


macisamuele:
@bxm156 thanks a lot for your solution ... it's great.
I would ask if you can map all the known standard http responses ... you can find them on the [IANA Standard](http://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml).

Maybe you could try to get all of them using this.
```bash
# Classes generation
curl http://www.iana.org/assignments/http-status-codes/http-status-codes-1.csv | \
    grep -vE "(Value,Description,Reference|Unassigned|Unused)" | \
    awk -F, '{ if($1>=400) { gsub(" ","", $2); printf "\n\nclass HTTP%s(HTTPClientError):\n        status = %s\n", $2, $1}}'
```

bxm156:
@macisamuele  sure thing

bravado/exception.py
```diff
@@ -53,10 +53,33 @@ def __str__(self):
...
+
 class HTTPClientError(HTTPError):
     """4xx responses."""
 
 
+class HTTPServerError(HTTPError):
+    """5xx responses."""
+
+
+# The follow are based on the HTTP Status Code Registery at
```
macisamuele: nit: Registery -> Registry

bxm156: Nooooooo!!!! I was so close to having no spelling errors :(

macisamuele:
@bxm156 Thanks a lot for your contribution.

PR #262, opened 2016-10-03
Update HTTPError to access status_code attribute directly.

### LINKED ISSUE #260: http exception by status code causes AttributeError: str has no status_code
See comments on:
https://github.com/Yelp/bravado/commit/87fa6813c7367a066eaa76db0dfad8483067aad4#diff-0a21d9158985f1ced07c9ef0ad7bb297

```
>       self.status_code = getattr(self.response, 'status_code')
E       AttributeError: 'str' object has no attribute 'status_code'
```

This may be bad testing code on our end.  Either way the `getattr` should be changed to just attribute access or to default to `None`

bxm156:
Update HTTPError to access status_code attribute directly. fixes #260

### DIFF
--- a/bravado/exception.py
+++ b/bravado/exception.py
@@ -40,7 +40,7 @@ class HTTPError(with_metaclass(HTTPErrorType, IOError)):
         self.response = response
         self.message = message
         self.swagger_result = swagger_result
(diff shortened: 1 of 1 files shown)


asottile:
#261

Issue #221, opened 2016-05-16
Make it super easy to wrap SwaggerClient to add new features


analogue:
This is in reference to #220 and #193.  Basically, make it easy to wrap the SwaggerClient, Resource, and Operation easily add new features like:
- OAuth2 support
- Passing headers on each service call

Tasks:
- Extract the relevant bits from the internal bravado_decorators library
- Update documentation with a sample usage of the wrappers

toabi:
That would be a pretty useful thing…

PR #264, opened 2016-10-21
Api key header

travisbeck:
I'm having a lot of success using bravado, but I find sending normal authentication headers annoying. You either have to send them with each request using _request_options or use a custom client which it isn't clear how to do from the docs. It feels to me like this is functionality that bravado should be shipping with, so I went ahead and added it.

This branch adds a 'param_in' argument to ApiKeyAuthenticator and RequestsClient.set_api_key that when sent with the value 'header' adds a header with the name of 'param_name'. I called this param_in to stay in the spirit of the 'in' field in swagger for api_key auth.

I'm not sure if there is already another way to do this that I missed, but if there is it is definitely not well documented.

Thanks!

### DIFF
--- a/CHANGELOG-MASTER.rst
+++ b/CHANGELOG-MASTER.rst
@@ -7,4 +7,5 @@ Changelog-Master
 
 X.X.X (XXXX-XX-XX)
 ------------------
+- Add API key authentication via header to RequestsClient
 - PLACEHOLDER (add before this line your modifications summary)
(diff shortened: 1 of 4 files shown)


travisbeck:
Also, this seems pretty relevant to https://github.com/Yelp/bravado/issues/221

wasbazi:
What is necessary to get this PR over the line? This would be very helpful for my code.

sjaensch:
Sorry for the delay in merging this, thanks @travisbeck for your contribution!

PR #270, opened 2016-12-16
fix TypeError - requests.Request does not accept timeout arg

lucagiovagnoli:
There is a TypeError bug when setting a timeout on client creation.
The timeout option in request_options should only be used by swaggerpy as [it is not an argument](https://github.com/Yelp/bravado/blob/500bdae0f53b68fa24074ac7b447cc9508c0fe6e/swaggerpy/http_client.py#L176) of the requests.[Request](http://docs.python-requests.org/en/master/api/#requests.Request) object.

This will now be possible 

```python
SwaggerClient.from_url(
  u'http://localhost/api-docs',
  request_options={
    'timeout': 15,
  }
)
```

### DIFF
--- a/swaggerpy/swagger_model.py
+++ b/swaggerpy/swagger_model.py
@@ -20,6 +20,8 @@ SWAGGER_VERSIONS = [u"1.2"]
 
 log = logging.getLogger(__name__)
 
+# max timeout for swagger client initialization
+SWAGGER_INIT_TIMEOUT = 5
 
 # TODO: replace with swagger_schema_validator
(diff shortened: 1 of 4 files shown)


swaggerpy/swagger_model.py
```diff
@@ -20,6 +20,8 @@
 
 log = logging.getLogger(__name__)
 
+# max timeout for swagger client initialization
+SWAGGER_INIT_TIMEOUT = 5
```
laucia: you are missing a blank line here

Issue #227, opened 2016-05-26
Should fido / twisted / requests be optional requirements?


asottile:
Or perhaps split out into separate packages?  To me it doesn't make too much sense to drag them along if they're not essential for functionality

laucia:
They provide the networking capabilities, and that is what bravado brings on top of bravado-core.
So yes I think they are required.

Unless we want bravado not to be usable out of the box ?

asottile:
but I can use bravado without using twisted or I can use bravado without using fido?

lucagiovagnoli:
Sure, you can use bravado without using Fido (and hence Twisted) if you select the RequestClient. Or you can use bravado without using RequestClient if you select FidoClient. Bravado is the swagger-validation layer, it can use any http client, and we currently provide 2 of them out of the box

asottile:
My suggestion is to split the specific implementations into other
packages.  If I'm just using, say, the twisted client then I really don't
care about fido, requests, and all their dependencies.  But currently
bravado requires me to install them.  If they're truly plug-and-play I see
no reason to put them in the core package.

torwald-sergesson:
I think, it will be fine to move specific clients into another boxes. For example, bravado-fido-client, bravado-requests-client etc. Today I've tried to install bravado on CentOS 7 with python 3.5.2. And it was failed because of problems with twisted. But I'm not going to use twisted in my application, as fido client. I would prefer to install one more library, then to fix problems with twisted by forking.

amilstead:
+1 to not including all of the implementation dependencies as part of Bravado's base installation. Separate packages isn't the only option, though -- maybe pick a default (perhaps `requests`) then include that as part of the main bravado dependency list and drop the others from the requirements file. The other implementations can stay in code but will raise `ImportError`s if they are used.

The `extras_require` feature from setuptools (documented [here](http://setuptools.readthedocs.io/en/latest/setuptools.html#new-and-changed-setup-keywords)) could be used to include other libraries dependent for the implementations that would now raise out on import.

asottile:
seems this is no longer an issue

PR #273, opened 2017-02-21
Fido as optional dependency

macisamuele:
This PR is a suggestion for issue #227.

The idea is to provide by default bravado without dependencies for fido client and installing all the dependencies only if explicitly required.

``pip install bravado`` will not install ``fido`` related dependencies, while ``pip install bravado[fido]`` will install them.

Unfortunately it is not possible, at least with the current implementation, to make bravado with fido independent from requests because of [this block](https://github.com/Yelp/bravado/blob/2e7edec76517386582ce111d48c5489dba73efa2/bravado/fido_client.py#L100-L113).

To avoid skipping all the integration tests in case of packaging without fido I have have mirrored ``fido_client_test.py`` to ``requests_client_test.py``.

Suggestions and improvements will be appreciated.

PS. I have removed python2.6 definitions from setup.py because we dropped support in 8.4.0 (September 2016).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,22 +1,16 @@
-
-- repo: git://github.com/pre-commit/pre-commit-hooks
-  sha: 5fe82b
-  hooks:
-    - id: autopep8-wrapper
-      args: ['-i', '--ignore=E309']
-    - id: check-json
-    - id: check-yaml
-    - id: debug-statements
-    - id: end-of-file-fixer
-    - id: flake8
-    - id: name-tests-test
-    - id: trailing-whitespace
-    - id: requirements-txt-fixer
-      files: 'requirements-dev.txt'
-
-# TODO: re-enable in bravado, disabled for now to avoid excessive merge
-# conflicts
-#- repo: git://github.com/FalconSocial/pre-commit-python-sorter
-#  sha: 1.0.1
-#  hooks:
-#    - id: python-import-sorter
+-   repo: git://github.com/pre-commit/pre-commit-hooks
+    sha: v0.7.1
+    hooks:
+    -   id: autopep8-wrapper
+        args:
+        - -i
+        - --ignore=E309,E501
+    -   id: check-json
+    -   id: check-yaml
+    -   id: debug-statements
+    -   id: end-of-file-fixer
+    -   id: flake8
+    -   id: name-tests-test
+    -   id: trailing-whitespace
+    -   id: requirements-txt-fixer
+        files: requirements-dev.txt
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -7,6 +7,7 @@
         - --ignore=E309,E501
     -   id: check-json
     -   id: check-yaml
+    -   id: fix-encoding-pragma
     -   id: debug-statements
     -   id: end-of-file-fixer
     -   id: flake8
@@ -14,3 +15,7 @@
     -   id: trailing-whitespace
     -   id: requirements-txt-fixer
         files: requirements-dev.txt
+-   repo: git://github.com/asottile/reorder_python_imports
+    sha: v0.3.1
+    hooks:
+    -   id: reorder-python-imports
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,25 @@
 language: python
+
+python:
+- "2.7"
+- "3.5"
+env:
+- TOX_FLAVOR=default
+- TOX_FLAVOR=fido
+
 matrix:
   include:
-  - python: "3.5"
-    env: TOX_ENV=py35
-env:
-- TOX_ENV=py27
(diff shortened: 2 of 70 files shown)


sjaensch:
@asottile could you take a look at this as well?

macisamuele:
@sjaensch I was forgetting to mention that after the initial review and eventual approval I have to update the CHANGELOG and README (adding examples of use with requests and fido)

.pre-commit-config.yaml
```diff
@@ -1,22 +1,16 @@
...
-    - id: trailing-whitespace
-    - id: requirements-txt-fixer
-      files: 'requirements-dev.txt'
-
-# TODO: re-enable in bravado, disabled for now to avoid excessive merge
-# conflicts
-#- repo: git://github.com/FalconSocial/pre-commit-python-sorter
-#  sha: 1.0.1
-#  hooks:
-#    - id: python-import-sorter
```
asottile: I'd suggest instead https://github.com/asottile/reorder_python_imports

.travis.yml
```diff
@@ -1,17 +1,25 @@
...
-- TOX_ENV=py27
-- TOX_ENV=cover
-- TOX_ENV=docs
-- TOX_ENV=flake8
+  - python: "2.7"
+    env: TOX_ENV=cover
+  - python: "2.7"
+    env: TOX_ENV=docs
+  - python: "2.7"
+    env: TOX_ENV=flake8
```
asottile: Instead of being cute, let's be explicit:

```yaml
matrix:
  include:
  - env: TOXENV=py27-default
  - env: TOXENV=py27-fido
  - env: TOXENV=py36-default
    python: python3.6
  - env: TOXENV=py36-fido
    python: python3.6
  - env: TOXENV=cover
  - env: TOXENV=docs
  - env: TOXENV=flake8
```

This is _much_ clearer what's going on :)

setup.py
```diff
@@ -8,6 +8,12 @@
 
 import bravado
 
+fido_client_dependencies = [
```
asottile: no need for unnecsesary indirection, just inline these.  It's already pretty obvious these are for "fido" since they are under the fido block.

Why do you need an explicit dependency on crochet / yelp-bytes, shouldn't fido be handling that?

tests/integration/conftest.py
```diff
@@ -0,0 +1,48 @@
+# -*- coding: utf-8 -*-
+
```
asottile: p3: if you add the pre-commit hook I suggested, this newline will go away :)

tests/integration/conftest.py
```diff
@@ -0,0 +1,48 @@
...
+@pytest.yield_fixture(scope='session')
+def threaded_http_server():
+    port = get_hopefully_free_port()
+    thread = threading.Thread(
+        target=bottle.run, kwargs={'host': 'localhost', 'port': port},
+    )
+    thread.daemon = True
+    thread.start()
+    time.sleep(2)
```
asottile: instead of flakey time.sleep, consider using urllib until it connects

tests/integration/conftest.py
```diff
@@ -0,0 +1,48 @@
...
+    s = socket.socket()
+    s.bind(('', 0))
+    port = s.getsockname()[1]
+    s.close()
+    return port
+
+
+@pytest.yield_fixture(scope='session')
+def threaded_http_server():
+    port = get_hopefully_free_port()
```
asottile: pip install ephemeral-port-reserve, and use `ephemeral_port_reserve.reserve`.  It's race-free if you use the port within 60 seconds

tox.ini
```diff
@@ -1,22 +1,32 @@
...
-envlist = py27, py35, flake8
+envlist = {py27,py35}-{default,fido}, flake8
 
 [testenv]
 deps =
     -rrequirements-dev.txt
+extras =
+    fido: fido
+setenv =
+    default: PYTEST_ADDOPTS=--ignore=tests/fido_client --ignore=tests/integration/fido_client_test.py
```
asottile: why not just always install fido, what is gained by splitting the two testenvs?

tests/integration/conftest.py
macisamuele: Actually we should add ``fix-encoding-pragma`` hook to the codebase since we're supporting python2 and python3 at the same time.

tox.ini
macisamuele: I would prefer to have separated tests because we should make sure that the library is working as expected even without the extra ``fido`` dependency.

tests/integration/conftest.py
macisamuele: 👍

.pre-commit-config.yaml
macisamuele: 👍

macisamuele:
Going to merge this on master!

PR #274, opened 2017-03-09
Enforce usage of new pip version

macisamuele:
This will solve/mitigate https://github.com/pypa/pip/issues/3903


asottile:
Isn't this too late to have any affect on testing?  the new pip won't be used to install libraries

PR #275, opened 2017-03-10
Fix test failures caused by swagger-spec-validator 2.0.3

sjaensch:
swagger-spec-validator ships with an updated Swagger 2.0 schema, which
causes two tests two fail since the 'required' field is missing from the
param definition in those tests. This fixes our internal build failure.

I've also made string quotes consistent in that file.


macisamuele:
@sjaensch  great catch ... 🚢

PR #276, opened 2017-03-10
Fix build if using older tox

sjaensch:
Older tox versions don't know about `extras`, let's use a different way to install them.


macisamuele:
🚢

Issue #286, opened 2017-06-07
Binary file upload results in UnicodeError


akanak:
For a given operation of type multipart/form-data and parameter of type file, an attempt to initiate a swagger request to upload/transfer a binary file results in the following stacktrace:

`
lib/python2.7/site-packages/bravado/http_future.py", line 70, in result
    inner_response = self.future.result(timeout=timeout)
 lib/python2.7/site-packages/bravado/requests_client.py", line 271, in result
    timeout=self.build_timeout(timeout))
 lib/python2.7/site-packages/requests/sessions.py", line 573, in send
    r = adapter.send(request, **kwargs)
lib/python2.7/site-packages/requests/adapters.py", line 370, in send
    timeout=timeout
  lib/python2.7/site-packages/requests/packages/urllib3/connectionpool.py", line 544, in urlopen
    body=body, headers=headers)
  lib/python2.7/site-packages/requests/packages/urllib3/connectionpool.py", line 349, in _make_request
    conn.request(method, url, **httplib_request_kw)
  File "/usr/lib/python2.7/httplib.py", line 1017, in request
    self._send_request(method, url, body, headers)
  File "/usr/lib/python2.7/httplib.py", line 1051, in _send_request
    self.endheaders(body)
  File "/usr/lib/python2.7/httplib.py", line 1013, in endheaders
    self._send_output(message_body)
  File "/usr/lib/python2.7/httplib.py", line 862, in _send_output
    msg += message_body
UnicodeDecodeError: 'ascii' codec can't decode byte 0xff in position 614: ordinal not in range(128)
`
When a Swagger endpoint is invoked, bravado client invokes the http request and returns a future. In doing so, it constructs the request from the http method, parameters and the headers.  These fields are merged to form a message that is then sent over the wire.  The http method being a unicode string, the file content (binary) is also translated to unicode resulting in the above error. 

The following code in bravado/client.py  may be fixed to convert http_method to str to resolve the error.
`  
def construct_request(....):
    ......
    url = operation.swagger_spec.api_url.rstrip('/') + operation.path_name
    request = {
        'method': operation.http_method.upper(),
        'url': url,
        'params': {},  # filled in downstream
        'headers': request_options.get('headers', {}),
    }
`

PR #287, opened 2017-06-07
Convert http method to str while constructing the request

akanak:
#286


macisamuele:
This is wired ... as far as I can tell we never had similar issues either on Python 2.7 and Python 3.5/6 .
Are you able to write a test that is able to reproduce the issue and check that this is failing with the old code and succeeding with the new one? It will be really helpful to limit future regression.
Thanks a lot in your bravado support.

bravado/client.py
```diff
@@ -280,7 +280,7 @@ def construct_request(operation, request_options, **op_kwargs):
     """
     url = operation.swagger_spec.api_url.rstrip('/') + operation.path_name
     request = {
-        'method': operation.http_method.upper(),
+        'method': str(operation.http_method.upper()),
```
sjaensch: str() does not convert to a binary string in Python 3 - as @macisamuele said a test for this would be great. `bytes()` is supported both in Python 2.7 and Python 3, but needs to be called differently depending on the Python version. You'll need to handle this a bit differently. What will do the trick is

    operation.http_method.upper().encode('ascii')

if the http_method is guaranteed to be str in Python 3 (it can be either str or unicode in Python 2).

macisamuele:
Hey @akanak could you apply the following diff to your repo? In this way we will have an additional Travis environment in which we will run tests with ``requests<2.8``
```diff
diff --git a/.travis.yml b/.travis.yml
index 05c18c3..7fe8465 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ matrix:
   include:
   - env: TOXENV=py27-default
   - env: TOXENV=py27-fido
+  - env: TOXENV=py27-fido-requests2dot8

   - env: TOXENV=py35-default
     python: "3.5"
diff --git a/tox.ini b/tox.ini
index c8d2324..297c393 100644
--- a/tox.ini
+++ b/tox.ini
@@ -1,10 +1,11 @@
 [tox]
-envlist = {py27,py35,py36}-{default,fido}, flake8
+envlist = {py27,py35,py36}-{default,fido}, {py27,py35,py36}-fido-requests2dot8, flake8

 [testenv]
 deps =
     -rrequirements-dev.txt
     fido: .[fido]
+    requests2dot8: requests<2.8
 setenv =
     default: PYTEST_ADDOPTS=--ignore=tests/fido_client --ignore=tests/integration/fido_client_test.py
 commands =
```

bravado/client.py
akanak: encode('ascii') converts a unicode string to bytes in python3, and this throws ConnectionError (possibly due to encoding assumed by the petstore server?). Added a test case with requests pinned to < 2.8, and it fails for py35 and py36 with the above code changes.

akanak:
Test passes with py27, py35 and py36.

PR #288, opened 2017-06-11
Injected headers are casted to strings

macisamuele:
As done in [bravado-core PR #115](https://github.com/Yelp/bravado-core/pull/115) we should cast headers injected via ``_request_options`` to string.
This is a workaround for [requests issue #3491](https://github.com/requests/requests/issues/3491). More context on bravado-core PR

### DIFF
--- a/bravado/client.py
+++ b/bravado/client.py
@@ -283,7 +283,9 @@ def construct_request(operation, request_options, **op_kwargs):
         'method': str(operation.http_method.upper()),
         'url': url,
         'params': {},  # filled in downstream
-        'headers': request_options.get('headers', {}),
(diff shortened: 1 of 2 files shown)


sjaensch:
@macisamuele I think we're converting the header values too soon. We've got internal reports that this change breaks validation for non-string parameters that are passed in the header. I think we need to do the stringifying in bravado-core.

macisamuele:
@sjaensch ... that's weird to me ... are we able to reproduce the issue in order to write a test that targets it?
We can "easily" moving the conversion to string of the headers just before the return statement (https://github.com/Yelp/bravado/pull/288/files#diff-b7921701c1024606d23a5ebb67b3a930R297)

PR #289, opened 2017-06-12
[SwaggerPy] Injected headers are casted to strings

macisamuele:
Context on #288  (same change from Swagger 2.0 client)

@sjaensch could we even release this new version ?


macisamuele:
@sjaensch sure ... let's release internally

PR #292, opened 2017-06-21
Make sure we throw the correct timeout exception when using the fido client

sjaensch:
You can pass two timeout values to bravado: One as part of the _request_options parameter when creating a future, and another one when calling result(). The timeout triggered by the _request_options parameter is handled by fido internally - it's a twisted timeout error.

The call to result() with a timeout parameter is unfortunately converted to a a crochet.TimeoutError  by crochet's eventloop ([here](https://github.com/itamarst/crochet/blob/8bba614f9f3d64c3741a43f48607e8ff6c5b50a1/crochet/_eventloop.py#L192)). Let's make sure we convert that exception to a fido HTTPTimeoutError for consistency. It's also what our internal retry mechanism expects.

### DIFF
--- a/bravado/fido_client.py
+++ b/bravado/fido_client.py
@@ -3,6 +3,7 @@ import logging
 
 import crochet
 import fido
+import fido.exceptions
 import requests
 from bravado_core.response import IncomingResponse
 from yelp_bytes import to_bytes
@@ -148,4 +149,8 @@ class FidoFutureAdapter(FutureAdapter):
             return self._eventual_result.wait(timeout=timeout)
         except crochet.TimeoutError:
             self._eventual_result.cancel()
-            raise
+            raise fido.exceptions.HTTPTimeoutError(
+                'Connection was closed by fido after blocking for '
+                'timeout={timeout} seconds waiting for the server to '
+                'send the response'.format(timeout=timeout)
+            )
--- a/bravado/fido_client.py
+++ b/bravado/fido_client.py
@@ -1,10 +1,12 @@
 # -*- coding: utf-8 -*-
 import logging
+import sys
 
 import crochet
 import fido
 import fido.exceptions
 import requests
+import six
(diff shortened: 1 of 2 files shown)


bravado/fido_client.py
```diff
@@ -148,4 +149,8 @@ def result(self, timeout=None):
             return self._eventual_result.wait(timeout=timeout)
         except crochet.TimeoutError:
             self._eventual_result.cancel()
-            raise
+            raise fido.exceptions.HTTPTimeoutError(
```
laucia: in python 2 we are going to loose the traceback here, are you sure this is OK ?

macisamuele: +1

sjaensch: I think it's fine - the original traceback doesn't really give any more info as to what's going on. It looks like this:

      File "<stdin>", line 1, in <module>
      File "...lib/python3.6/site-packages/bravado/http_future.py", line 70, in result
        inner_response = self.future.result(timeout=timeout)
      File "...lib/python3.6/site-packages/bravado/fido_client.py", line 147, in result
        return self._eventual_result.wait(timeout=timeout)
      File "...lib/python3.6/site-packages/crochet/_eventloop.py", line 245, in wait
        result = self._result(timeout)
      File "...lib/python3.6/site-packages/crochet/_eventloop.py", line 209, in _result
        raise TimeoutError()
    crochet._eventloop.TimeoutError

I think having a consistent exception type is far more important - the traceback only really matters when debugging internal issues (i.e. when developing bravado / fido).

laucia: Alright: just pointing that this was never about not doing it, just doing it as a reraise.

Anyway python 3 ftw so let's go

sjaensch: Ah got it, you were going for six.reraise. :) I updated the code.

Issue #222, opened 2016-05-19
Gracefully reraise JSONDecodeError errors


askedrelic:
We've run into this issue several times, where a swagger service has crashed and is returning HTML 500s, which then causes our client to raise JSONDecodeError.

These are pretty ugly and hard to debug; you can't see what the not-json response is.

I feel like the bravado client should handle this case and reraise as a Bravado custom Exception, with the crash context/original response available.

Thoughts? I will leave this open, but don't have immediate plans to work on this.

Here is a simplified crash:
```
File "virtualenv_run/local/lib/python2.7/site-packages/swaggerpy/client.py", line 159, in get_client
return cache(*args, **kwargs)
File "virtualenv_run/local/lib/python2.7/site-packages/swaggerpy/client.py", line 118, in __call__
self.build_client(*args, **kwargs), ttl)
File "virtualenv_run/local/lib/python2.7/site-packages/swaggerpy/client.py", line 124, in build_client
return SwaggerClient.from_url(api_docs, *args, **kwargs)
File "virtualenv_run/local/lib/python2.7/site-packages/swaggerpy/client.py", line 326, in from_url
load_resource_listing(url, http_client, None, request_options),
File "virtualenv_run/local/lib/python2.7/site-packages/swaggerpy/swagger_model.py", line 196, in load_resource_listing
add_api_docs(resource_listing)
File "virtualenv_run/local/lib/python2.7/site-packages/swaggerpy/swagger_model.py", line 184, in add_api_docs
api['api_declaration'] = eventual.wait(timeout=timeout).json()
File "virtualenv_run/local/lib/python2.7/site-packages/requests/models.py", line 802, in json
return json.loads(self.text, **kwargs)
File "virtualenv_run/local/lib/python2.7/site-packages/simplejson/__init__.py", line 505, in loads
return _default_decoder.decode(s)
File "virtualenv_run/local/lib/python2.7/site-packages/simplejson/decoder.py", line 370, in decode
obj, end = self.raw_decode(s)
File "virtualenv_run/local/lib/python2.7/site-packages/simplejson/decoder.py", line 400, in raw_decode
return self.scan_once(s, idx=_w(s, idx).end())
JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

laucia:
This was probably fixed by https://github.com/Yelp/bravado-core/pull/97

sjaensch:
[...]

PR #297, opened 2017-06-27
raise on 400 errors

openfly:
otherwise you get json decode errors ( which is not very informative )


sjaensch:
This change breaks tests and has quite far-reaching consequences, although I agree with you that we should handle this case better. @openfly if you're interested in digging into this deeper check out issue #222 for a start.

A change to handle this better should come with some tests that exercise the new behavior and make sure it works correctly.

openfly:
gonna be honest.. i have bigger fish to fry atm.

PR #305, opened 2017-07-26
Make sure headers are unicode when using the fido client

sjaensch:
This change makes sure header names and values are the same as with the requests client when using the fido client. twisted (the underlying library fido uses) returns headers as a dictionary of binary string keys and lists of binary string values. They are now both unicode strings, and the header names can be accessed in a case-insensitive manner.

Due to our almost non-existing integration test coverage for the fido client we can't really rely on tests to tell us if this change breaks anything. If I have time I'll try to work on that later.

bravado-core 4.8.0 also broke a bunch of bravado tests, the one line change in conftest.py fixes them.

### DIFF
--- a/CHANGELOG-MASTER.rst
+++ b/CHANGELOG-MASTER.rst
@@ -5,3 +5,7 @@ Changelog-Master
 
 *The content will be used to build the Changelog of the new bravado release.*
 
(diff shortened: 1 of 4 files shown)


shrayus:
hi, we're waiting on this change to get merged is there a timeline for the next action?

PR #310, opened 2017-08-24
Require requests >= 2

nattofriends:
RequestsClient calls requests Session.prepare_request (among others?) which is only present in 2.0.0 and above.


laucia:
Thanks for this

Issue #307, opened 2017-07-31
Resources created when tags have spaces are inaccessible


ritwik:
This issue is a mix of Bravado / Bravado core so I'm filing here but happy to move to Bravado core if that is more appropriate.

# Summary:
When a resource has tags which have spaces, e.g "contact activites", its operations become inaccessible in the generated swagger client, because they must be accessed with the space (which isn't possible in python).

# Background:
We use bravado to create a client for this swagger spec: http://api.cordial.io/docs/v1/api-docs.php
In that spec, there are some resources with two word names (e.g `contactactivities`, `messagetransports`). In that spec, the tags for these have spaces in them (`contact activities`, `message transports`). See snippets below:

```json
"/v1/contactactivities":{
            "get":{
                "summary": "Get Activity list",
                "operationId": "getActivityList",
                "tags":[
                    "contact activities"
                ],
  ...
```

```json
"/v1/messagetransports/{id}":{
            "get": {
                "summary": "Get a transport's properties",
                "operationId": "getsingletransport",
                "tags":[
                    "message transports"
                ],
  ...
```

When we build a swagger client from this spec using bravado, we can't access those resouces:
```python
client = SwaggerClient.from_spec(
    CORDIAL_SPEC,
    http_client=http_client,
)
client.contactactivities.getActivityList() # Raises exception.
client.contact_activities.getActivityList() # Raises exception.

# Example exception (with some noise removed):
# ...
# File "../sales/coupons.py", line 256, in _send_to_cordial
#    client.contactactivities.addActivity(a="gotCoupon", email=email,
#  File "../lib/thirdparty/bravado/client.py", line 161, in __getattr__
#    .format(item, ', '.join(dir(self))))
# AttributeError: Resource contactactivities not found. Available resources: account attributes, account lists, automation templates, batch messages, contact activities, contact exports, contact imports, contacts, jobs, message content includes, message transports, orders, productimports, products, supplements
```

The exceptions happen because (from what I understand so far):
- `client.<anything>` works by overriding `__getattr__(self, attr)` to look up `attr` inside the resources dictionary
- bravado_core builds the resources dictionary from the tags if they exist. This means that for this spec, the dictionary has: `contact activities` and `message transports` as keys. 
- There's no way to access those without manually calling `__getattr__`. e.g It works if I do: `client.__getattr__("contact activities")`, but it seems wrong to directly access that as a user of the client.

Unfortunately I don't control the spec, so we can't just add more tags that are accessible.

[...]

PR #320, opened 2017-10-11
Add SwaggerClient.get_resource method

jakul:
Fixes https://github.com/Yelp/bravado/issues/307

### DIFF
--- a/bravado/client.py
+++ b/bravado/client.py
@@ -148,10 +148,7 @@ class SwaggerClient(object):
     def get_model(self, model_name):
         return self.swagger_spec.definitions[model_name]
 
-    def __repr__(self):
-        return u"%s(%s)" % (self.__class__.__name__, self.swagger_spec.api_url)
-
-    def __getattr__(self, item):
+    def get_resource(self, item):
         """
         :param item: name of the resource to return
         :return: :class:`Resource`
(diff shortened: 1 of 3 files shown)


tests/client/SwaggerClient/getattr_test.py
```diff
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 import pytest
 
-from bravado.client import ResourceDecorator
+from bravado.client import ResourceDecorator, SwaggerClient
```
sjaensch: Please install the pre-commit hooks (`tox -e pre-commit install`) and run them (`.tox/pre-commit/bin/pre-commit run --all-files`). This will happen automatically for the files you changed for every git commit after you installed them.

bravado/client.py
```diff
@@ -148,10 +148,7 @@ def from_spec(cls, spec_dict, origin_url=None, http_client=None,
     def get_model(self, model_name):
         return self.swagger_spec.definitions[model_name]
 
-    def __repr__(self):
-        return u"%s(%s)" % (self.__class__.__name__, self.swagger_spec.api_url)
-
-    def __getattr__(self, item):
+    def get_resource(self, item):
```
sjaensch: This could in theory break users that have a tag of "get_resource". Obviously we didn't worry about this back when this class was written, as we also have other methods that don't start with an underscore. That said, I wonder if we should start following that practice now. It's not consistent (which is not great), but still seems like the right thing to do.

Could you please rename the method to "_get_resource"? We do this for the Model class as well.

tests/client/SwaggerClient/getattr_test.py
jakul: Done.

bravado/client.py
jakul: Done.

jakul:
@sjaensch I've actioned all the changes you requested.

PR #321, opened 2017-10-14
Unify timeout errors

macisamuele:
The goal of this PR is to provide a standard way of catching timeout error thrown by the different HTTP clients implementation.

It exposes a new exception ``BravadoTimeoutError`` which will be used as base class of the different HTTP timeout errors.

The _magic_ is provided by ``raraise_timeout_errors`` decorator.
The decorator ensures that whenever ``HttpFuture.result`` is called it will reraise timeout errors (defined by ``FutureAdapter.timeout_errors`` attribute) that could be catched as ``BravadoTimeoutError`` exceptions.

To make this change transparent and compatible with third-party HTTP clients (like [bravado-asyncio](https://github.com/sjaensch/bravado-asyncio)) the decorator does no assumption about the presence of ``timeout_errors`` attribute in the future adapter class.

While preparing this PR I have also extracted all the base logic for the integration tests to a common file (``conftest.py``). This should simplify, via another PR, to expose them to third-party clients implementations.

### DIFF
--- a/bravado/exception.py
+++ b/bravado/exception.py
@@ -1,6 +1,12 @@
 # -*- coding: utf-8 -*-
-from six import with_metaclass
+try:
+    from builtins import TimeoutError as base_exception
+except ImportError:
+    # TimeoutError was introduced in python 3.3+
+    base_exception = Exception
 
+
+from six import with_metaclass
 
 # Dictionary of HTTP status codes to exception classes
 status_map = {}
@@ -230,3 +236,7 @@ class HTTPNotExtended(HTTPServerError):
 
 class HTTPNetworkAuthenticationRequired(HTTPServerError):
     status_code = 511
+
+
+class BravadoTimeoutError(base_exception):
+    pass
--- a/bravado/http_future.py
+++ b/bravado/http_future.py
@@ -35,7 +35,7 @@ class FutureAdapter(object):
         )
 
 
-def raraise_timeout_errors(func):
+def reraise_errors(func):
     @wraps(func)
     def wrapper(self, *args, **kwargs):
         timeout_errors = tuple(getattr(self.future, 'timeout_errors', None) or ())
@@ -95,7 +95,7 @@ class HttpFuture(object):
         self.response_callbacks = response_callbacks or []
         self.also_return_response = also_return_response
 
-    @raraise_timeout_errors
+    @reraise_errors
     def result(self, timeout=None):
         """Blocking call to wait for the HTTP response.
 
--- a/bravado/http_future.py
+++ b/bravado/http_future.py
@@ -1,10 +1,12 @@
 # -*- coding: utf-8 -*-
 import sys
+from functools import wraps
 
 import bravado_core
 import six
 from bravado_core.exception import MatchingResponseNotFound
 
+from bravado.exception import BravadoTimeoutError
 from bravado.exception import make_http_exception
 
 
@@ -18,6 +20,9 @@ class FutureAdapter(object):
 
     """
 
+    # Make sure to define the timeout errors associated with your http client
+    timeout_errors = []
+
     def result(self, timeout=None):
         """
         Must implement a result method which blocks on result retrieval.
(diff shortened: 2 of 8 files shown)


bravado/exception.py
```diff
@@ -230,3 +236,7 @@ class HTTPNotExtended(HTTPServerError):
 
 class HTTPNetworkAuthenticationRequired(HTTPServerError):
     status_code = 511
+
+
+class BravadoTimeoutError(base_exception):
```
sjaensch: This is a backwards-incompatible change, and changing the type of timeout exception fido threw when calling `result()` caused quite a bit of pain until we found a backwards-compatible solution. Maybe we could do the same here?

1. Define client-specific timeout exceptions like RequestsTimeoutError, FidoTimeoutError, AsyncioTimeoutError
1. Make them inherit from both BravadoTimeoutError and their "native" timeout exception, like fido.exceptions.HTTPTimeoutError, requests.exceptions.ReadTimeout or aiohttp.ServerTimeoutError
1. Do the same thing for connection errors (we have the same issue there)

While it'd be ideal if we could implement this unification only in bravado, I'm perfectly fine with extending the API contract between bravado and the HTTP client to require the client to define their own connection and timeout exceptions that need to be subclasses of bravado ones.

I also think there is value in requiring HTTP clients to behave like that: We can be certain that the bravado library with this change will throw consistent timeout and connection exceptions. We'll just require new fido and bravado-asyncio versions to be used with it. I.e. I probably wouldn't make @reraise_timeout_errors fail gracefully.

bravado/http_future.py
```diff
@@ -30,6 +35,37 @@ def result(self, timeout=None):
         )
 
 
+def raraise_timeout_errors(func):
```
sjaensch: r**e**raise_timeout_errors

bravado/exception.py
macisamuele: I don't see how this change could be back incompatible.
``raraise_timeout_errors`` decorator makes sure to raise exceptions dynamically generated if not defined in ``FutureAdapter.__timeout_error_type`` that are extending the timeout exceptions and ``BravadoTimeoutException`` ([Reference](https://github.com/Yelp/bravado/pull/321/files#diff-dfa77bc847cf31d43435498de902fe9bR47)).
This means that you can continue to catch timeout exceptions how you were used to (using HTTP client specific timeout exceptions) and using the new base class for bravado timeout exceptions.

About connections errors, I was thinking to unify this type of errors too but I was having issues in writing a valid integration test for fido client so I decided to postpone it in a separate PR

tests/integration/conftest.py
```diff
@@ -51,9 +60,108 @@ def wait_unit_service_starts(url, timeout=10):
...
     thread.daemon = True
     thread.start()
-    wait_unit_service_starts('http://localhost:{port}'.format(port=port))
-    yield port
+    server_address = 'http://localhost:{port}'.format(port=port)
+    wait_unit_service_starts(server_address)
+    yield server_address
+
+
+class BaseClientIntegrationTestClass:
```
sjaensch: I'm not too happy with the design decision to put a significant number of tests in conftest.py, to be executed when this class is inherited. We should choose a different design instead and make use of pytest fixtures and probably get rid of the class approach altogether.

I don't think we necessarily need to separate the test files by HTTP client used. If we want to run a set of tests for all clients then let's just put it in the same file, which should help us get rid of the inheritance approach. It also enables us to do client-specific tests like you do in `TestServerRequestsClientFake` without having to rerun all of the common tests.

bravado/exception.py
sjaensch: Sounds good, I didn't look closely enough at the decorator implementation initially.

bravado/http_future.py
macisamuele: I'm renaming it to a more general ``reraise_errors`` so we could use the same decorator for handling connection errors (not done by this PR)

tests/integration/conftest.py
macisamuele: I guess that the original idea of having a base test class (previously ``TestServerRequestsClient`` was used as base) was to make sure that requests and fido have to pass the same tests.
I moved the base class definition in ``conftest`` in order to simplify the process of making this test class available for other libraries like bravado-asyncio

macisamuele:
I restored the base test class into ``requests_client_test.py`` as it originally was.
We will think a better approach to expose integration tests to external HTTP clients implementations

PR #324, opened 2017-11-03
Fixes #323 - Adds raw_bytes to IncomingResponse for msgpack support

### LINKED ISSUE #323: Add raw_bytes attribute to RequestClient and FidoClient
Once https://github.com/Yelp/bravado-core/issues/214 is completed, all classes using the `IncomingResponse` will be required to have a new attribute, `raw_bytes` defined. Responses in RequestClient and FidoClient will need to but updated for this change

AlstonLin:
See https://github.com/Yelp/bravado-core/pull/216 for context

### Testing
Ran `make test` and made sure there were no failures (only xfailures shown)

### DIFF
--- a/bravado/fido_client.py
+++ b/bravado/fido_client.py
@@ -38,6 +38,10 @@ class FidoResponseAdapter(IncomingResponse):
         return self._delegate.body
 
     @property
+    def raw_bytes(self):
+        return self._delegate.body
+
+    @property
     def reason(self):
         return self._delegate.reason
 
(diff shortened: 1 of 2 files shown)


bravado/fido_client.py
```diff
@@ -38,6 +38,10 @@ def text(self):
         return self._delegate.body
 
     @property
+    def raw_bytes(self):
+        return self._delegate.body
```
sjaensch: This seems to be a limitation of fido - the text property should return unicode, the raw_bytes property bytes. I'm assuming self._delegate.body is bytes, so this should be fine (and we should fix `text` above).

PR #334, opened 2017-12-04
Add a _use_msgpack option

### LINKED ISSUE #333: Add a _use_msgpack option
https://github.com/Yelp/bravado-core/pull/216 added support for msgpack, but to use it with bravado the following is required

```
result = swagger_client.tag.operation(
    body=your_body,
    _request_options={"headers": {
        "Accept": "application/msgpack"
    }},
).result()
```

instead, the header adding should happen behind the scenes, and there would be a simple flag that a user would enable to add it instead, so it would look something like

```
result = swagger_client.tag.operation(
    body=your_body,
    _use_msgpack=True,
).result()
```

AlstonLin:
Fixes #333 - Adds a _use_msgpack option for users to more easily specify that they want to request msgpack content and add the Accept header

## Testing (Manual)
- Added a print line in `bravado.requests_client.RequestsResponseAdapter.raw_bytes`
- In a `python` shell, imported `bravado` and created a client to a spec and server that would produce a `application/msgpack` Content-Type when the `Accept:` header is received, and JSON in all other cases
- Used the client to request a endpoint with no `_use_msgpack` specified and verified that JSON was printed from `raw_bytes` and the correct JSON was returned from `result()`
- Used the client to request a endpoint with `_use_msgpack=True` and verified that msgpack binary was printed and the correct JSON was returned from `result()`

(Note that all tests in `tests/client_test.py` xfailed, so I could not add a test to it and have it run properly)

### DIFF
--- a/bravado/client.py
+++ b/bravado/client.py
@@ -275,7 +275,7 @@ class CallableOperation(object):
             also_return_response=also_return_response)
 
 
-def construct_request(operation, request_options, **op_kwargs):
+def construct_request(operation, request_options, _use_msgpack=False, **op_kwargs):
     """Construct the outgoing request dict.
 
     :type operation: :class:`bravado_core.operation.Operation`
@@ -293,6 +293,9 @@ def construct_request(operation, request_options, **op_kwargs):
         'params': {},  # filled in downstream
         'headers': request_options.get('headers', {}),
     }
+    # Adds Accept header to request for msgpack response if specified
+    if _use_msgpack:
+        request['headers']['Accept'] = 'application/msgpack'
 
     # Copy over optional request options
     for request_option in ('connect_timeout', 'timeout'):


bravado/client.py
```diff
@@ -293,6 +293,9 @@ def construct_request(operation, request_options, **op_kwargs):
         'params': {},  # filled in downstream
         'headers': request_options.get('headers', {}),
     }
+    # Adds Accept header to request for msgpack response if specified
+    if _use_msgpack:
```
macisamuele: Please pass this flag via ``request_options`` as done for [``also_return_response``](https://github.com/Yelp/bravado/blob/2084666dba9f3407a2dbebfb96387a60b00b4e0c/bravado/client.py#L266).

AlstonLin:
@macisamuele is this ready to be merged?

PR #337, opened 2017-12-07
Use ReadTimeout as base timeout exception for the requests client

sjaensch:
This fixes issues with code that catches requests.exceptions.ReadTimeout, which we broke by using requests.exceptions.Timeout as base class.

### DIFF
--- a/bravado/requests_client.py
+++ b/bravado/requests_client.py
@@ -3,6 +3,7 @@ import logging
 
 import requests
 import requests.auth
+import requests.exceptions
 import six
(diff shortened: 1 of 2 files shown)


ngaya-ll:
What about `requests.exceptions.ConnectTimeout`?

macisamuele:
@ngaya-ll : connection errors are not integrated yet because we didn't find a reliable way of testing those type of errors with fido client; but definitely our plan is to "generalize" also those type of errors (https://github.com/Yelp/bravado/pull/346)

Issue #339, opened 2017-12-18
umsgpack is not a dependency


jessemyers:
Nonetheless, it is imported in `bravado/http_future.py`

```
>>> import bravado.http_future
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/jessemyers/code/opensource/bravado/bravado/http_future.py", line 6, in <module>
    import umsgpack
ModuleNotFoundError: No module named 'umsgpack'
```

This is breaking.

jessemyers:
Supeceded by #341

PR #336, opened 2017-12-06
Response handling is performed in bravado

macisamuele:
The objective of this PR is to let ``bravado`` to perform all the handling of the requests and responses and use ``bravado-core`` for validating, marsharling and unmarshaling swagger objects.

Currently ``bravado-core`` dependencies and complexity is growing in order to support new features on HTTP handling.
``bravado-core`` was designed to have the basic handling of _swagger objects_ and ideally it should have no knowledge of HTTP requests, content type/compression handling, etc.

This change was inspired from @sjaensch effort on [``aiobravado``](https://github.com/sjaensch/aiobravado) 😄 

**NOTE**: this is a WIP ... there are no tests yet for ``unmarshal_response_inner``

@AlstonLin : FYI

PR #340, opened 2017-12-18
#339: Make umsgpack an extra and handle its absence

jessemyers:
The addition of `umsgpack` is a breaking change. It should be handled more gracefully:

 -  Catch the import error
 -  Raise an exception if needed but not installed
 -  Add an extra dependency

### DIFF
--- a/bravado/http_future.py
+++ b/bravado/http_future.py
@@ -3,7 +3,11 @@ import sys
 from functools import wraps
 
 import six
-import umsgpack
+try:
+    from umsgpack import unpackb
+except ImportError:
+    unpackb = None
+
 from bravado_core.content_type import APP_JSON
 from bravado_core.content_type import APP_MSGPACK
 from bravado_core.exception import MatchingResponseNotFound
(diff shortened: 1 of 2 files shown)


macisamuele:
Hey @jessemyers thanks a lot for your support.

The import error was caused by #336 which added ``msgpack-python`` dependency but imports ``umsgpack``. Tests were green because ``bravado-core`` was used to depend on ``umsgpack``.

I would suggest to use a simpler approach in fixing this specific issue. My suggestion would be to fix the import on ``http_future.py`` by importing from ``msgpack`` instead from ``umsgpack``.

jessemyers:
OK, will do.

jessemyers:
@macisamuele I could use some advise here.

I did an attempt to swap in `msgpack`, but it doesn't pass tests. See #341 

My best guess is that the two libraries don't provide a compatible interface. (The failing tests are getting binary content, not text content).

I don't know the `msgpack` libraries except as rumor. My motivation for the first PR was to unblock my team, which isn't using it. I'd rather not have to become an expert on the libraries to get there.

PR #341, opened 2017-12-18
Using umsgpack breaks because it is not a dependency

jessemyers:
Per feedback from #340, remove explicit `umsgpack` usage


macisamuele:
According to the use of msgpsck in bravado the two libraries are compatible, the test failure is related to the import of umsgpack for integration tests (tests/integration/requests_client_test.py).
Please fix that import too

macisamuele:
@jessemyers thanks a lot for your effort :)
I'm merging this on master so you and your team could use it.
A new version of bravado will be released tomorrow.

PR #346, opened 2018-01-18
Add twisted errors to list of Fido timeouts

### LINKED ISSUE #344: Not all Fido timeout errors are caught
There are some twisted errors that occur with low request timeout (the timeout passed in `_request_options`). The ones i've seen are `twisted.internet.error.DNSLookupError` and `twisted.internet.error.ConnectingCancelledError`. These seem like more general errors, but I think they occur because of the fact that the request timeout cancels the twisted deferred.

Either way, if you set the request timeout low, you'll get these, which should be reraised.

The values that I needed to set for timeout to get these errors were:
`0.001` (1ms) for `twisted.internet.error.ConnectingCancelledError`
and
`0.0001` (0.1ms) for `twisted.internet.error.DNSLookupError` - Note: 0 also works

You may ask why anyone in their right mind would set these values so low, but can easily happen when programmatically setting the timeout.

bplotnick:
This fixes #344 by adding twisted.internet.error.DNSLookupError and twisted.internet.error.ConnectingCancelledError to the list of timeout errors.

As mentioned in the issues, these seem like more general than just timeouts, but I can reliably reproduce them by setting low request timeouts, so they should probably be caught somehow.

### DIFF
--- a/bravado/fido_client.py
+++ b/bravado/fido_client.py
@@ -8,6 +8,7 @@ import fido.exceptions
 import requests
 import requests.structures
 import six
+import twisted
 from bravado_core.response import IncomingResponse
 from yelp_bytes import to_bytes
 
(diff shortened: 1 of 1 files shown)


macisamuele:
@bplotnick sorry for the late reply
I'm personally a bit against this change because the errors that you're adding to ``timeout_error``s are more like connection errors.
I planned to add the same handling  for connection errors (defining ``BravadoConnectionErrors``) but I was not able to have reliable integration tests.

bplotnick:
@macisamuele Yeah I am not too happy with this either because the errors could probably be caused by other issues. However, I'm not entirely sure what the alternative is. Could we put some sort of "reasonable" minimum timeout? Is that going to be reliable?

PR #353, opened 2018-03-05
Clarify that this tool only supports OpenAPI v2.0

DStape:
OpenAPI 3.0 was recently released, however Bravado does not currently
support this.


MrLoh:
Is bravado still mainainted? It's 6 years later in OAS 3 is still not supported it seems.

Issue #352, opened 2018-02-19
Handling next and previous pagination of results


bdowling:
I'm looking for an example of dealing with pagination in the case where the API provides a next/previous URL for you to follow.

It is not clear to me if there is a streamlined way of handling this in Bravado already, but I was attempting to craft a wrapper class that would handle this in my schema.  If there is and someone can point me to the light I'd be happy to write some additional docs around this useful case.

I have done it initially for a single operation, but I am trying so see if there is an easy way to do this for all operations more generically.

Is there a way to pass back the URL and have it process it per the spec/model as defined?  What I did in my test was parse out the params (cursor) in the URL and craft a new operation request, but it feels like there should be a simpler shortcut here.

I'm sure this has been encountered, so any input you can provide would be helpful.

The latest 3.0 OpenAPI specs try to address this with [links](https://swagger.io/docs/specification/links/), wondering if a `x-links` feature may make this suit the generic case (where other URLs to other Models in the API also are returned).

bdowling:
Thoughts on this?

bdowling:
@macisamuele @sjaensch have any input on this?  Seems like a common need, lots of APIs are designed to return URL refs back into the API, next/prev is just another case of that.  How can one generically map those URLs back into the swagger client so you can fetch the model they point to?  There is a convert_path_to_resource method, but this doesn't go deeper into the operation or path parameters embedded in the URL.

This seems common enough that their could be generic support for handling this?

bdowling:
I came up with this, feels like I'm doing something that there should already be an easier way to do.  Ideally I was thinking that with the `format: url` and `x-operationId: some_op` that I added into my schema, I could return back a custom type for the URLs in the response that could have a `next.fetch()` function.  Looks like currently `unmarshall_schema_object` isn't passed enough info though to re-implement what I did here in that context.

```python
    def fetch_op_from_url(self, data, key, req, res):
        # lookup the last op response_spec
        # check the content_spec for x-operationId on key
        # call the model_op on the url
        url = data[key]
        op = req.operation
        response_spec = get_response_spec(status_code=res.status_code, op=op)
        content_spec = op.swagger_spec.deref(response_spec['schema'])

        model_op_id = content_spec['properties'][key].get('x-operationId', op.operation_id)
        if model_op_id == op.operation_id:
            # short circuit, common for next/prev links, etc
            model_op = op
        else:
            # Find it on another resource, surprised there is no single dict
            # for operationId since it is required to be unique across spec
[...]

PR #357, opened 2018-04-23
Add fetch_link to make it much easier to follow linked URLs within an API

bdowling:
This is an attempt at answering the question I had in #352

By simply adding `x-operationId: [linkedOperation]` in your spec to URLs that are returned from the API that are internal links, and adding `add_fetch_link: True` to your config, those URLs will be returned in the model as `FetchLink` objects that can be called to retrieve the named `[linkedOperation]' on the given URL (e.g. obtain the http_request on that url and operationId, as if a CallableOperation had been called).

I think this is fairly straightforward.  The only item I question is how I added the backref to `swagger_spec._client`, mainly so I could access my new SwaggerClient._get_operation from the context of http_future where it seems I only had access to the operation spec.  If there was some other way around this, please let me know.  Alternatively _get_operation could be added to bravado-core instead I suppose.  I just felt it was similar in vein to _get_resource that is in this module.

Without the new config option, there is no changed default behaviors.

I suppose I need to add some tests for this to be accepted, but figured I'd throw it out here to get some feedback.

### DIFF
--- a/bravado/client.py
+++ b/bravado/client.py
@@ -72,6 +72,8 @@ class SwaggerClient(object):
     def __init__(self, swagger_spec, also_return_response=False):
         self.__also_return_response = also_return_response
         self.swagger_spec = swagger_spec
+        # provide a backref to the client from the spec
+        self.swagger_spec._client = self
 
     @classmethod
     def from_url(cls, spec_url, http_client=None, request_headers=None, config=None):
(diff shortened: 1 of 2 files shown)


bdowling:
Example usage:

```python

while 'next' in data:
    url = data['next']
    if url is None:
        break
    print("URI: ", repr(url))
    nextr = url()
    data,res = nextr.result()
```

```yaml
  MultipleInstrumentData:
    type: object
    properties:
      count:
        type: integer
        format: int32
      next:
        type: string
        format: url
        x-operationId: 'instruments'
        x-nullable: true
      previous:
        type: string
        format: url
        x-operationId: 'instruments'
        x-nullable: true
      results:
        type: array
 ....
```

PR #358, opened 2018-04-24
Implement the new HTTPFuture.response method with fallback results

sjaensch:
This is a first rough PR so people get the idea. I just modified a few integration tests to showcase that the implementation works (at least as far as replacing calls to `result()` goes).

I'll add more tests tomorrow unless we decide on going a different route. I'm also more than prepared to change names if you have better suggestions. :)

### DIFF
--- a/bravado/http_future.py
+++ b/bravado/http_future.py
@@ -13,7 +13,15 @@ from msgpack import unpackb
 
 from bravado.config_defaults import REQUEST_OPTIONS_DEFAULTS
 from bravado.exception import BravadoTimeoutError
+from bravado.exception import HTTPServerError
 from bravado.exception import make_http_exception
+from bravado.response import BravadoResponse
+
+
+FALLBACK_EXCEPTIONS = (
+    BravadoTimeoutError,
+    HTTPServerError,
+)
 
 
 class FutureAdapter(object):
@@ -101,9 +109,50 @@ class HttpFuture(object):
         self.response_callbacks = response_callbacks or REQUEST_OPTIONS_DEFAULTS['response_callbacks']
         self.also_return_response = also_return_response
 
-    @reraise_errors
-    def result(self, timeout=None):
+    def response(self, timeout=None, fallback_response=None, exceptions_to_catch=FALLBACK_EXCEPTIONS):
         """Blocking call to wait for the HTTP response.
+
+        :param timeout: Number of seconds to wait for a response. Defaults to
+            None which means wait indefinitely.
+        :type timeout: float
+        :param fallback_response: callable that accepts an exception as argument and returns the
+            swagger result to use in case of errors
+        :type fallback_response: callable that takes an exception and returns a fallback swagger result
+        :param exceptions_to_catch: Exception classes to catch and call `fallback_response`
+            with. Has no effect if `fallback_response` is not provided. By default, `fallback_response`
+            will be called for read timeout and server errors (HTTP 5XX).
+        :type exceptions_to_catch: List/Tuple of Exception classes.
+        :return: A BravadoResponse instance containing the swagger result and response metadata.
+        """
+        swagger_result = None
+        try:
+            incoming_response = self._get_incoming_response(timeout)
+            if self.operation is not None:
+                unmarshal_response(
+                    incoming_response,
+                    self.operation,
+                    self.response_callbacks,
+                )
+
+                swagger_result = incoming_response.swagger_result
+            elif incoming_response.status_code >= 300:
+                raise make_http_exception(response=incoming_response)
+        except exceptions_to_catch as e:
+            if fallback_response:
+                swagger_result = fallback_response(e)
+                incoming_response = None
+            else:
+                raise
+
+        return BravadoResponse(
+            result=swagger_result,
+            incoming_response=incoming_response,
+        )
+
+    def result(self, timeout=None):
+        """DEPRECATED: please use the `response()` method instead.
+
+        Blocking call to wait for the HTTP response.
 
         :param timeout: Number of seconds to wait for a response. Defaults to
             None which means wait indefinitely.
@@ -111,8 +160,7 @@ class HttpFuture(object):
         :return: Depends on the value of also_return_response sent in
             to the constructor.
         """
-        inner_response = self.future.result(timeout=timeout)
-        incoming_response = self.response_adapter(inner_response)
+        incoming_response = self._get_incoming_response(timeout)
 
         if self.operation is not None:
             unmarshal_response(
@@ -129,6 +177,11 @@ class HttpFuture(object):
             return incoming_response
 
         raise make_http_exception(response=incoming_response)
+
+    @reraise_errors
+    def _get_incoming_response(self, timeout=None):
+        inner_response = self.future.result(timeout=timeout)
+        return self.response_adapter(inner_response)
 
 
 def unmarshal_response(incoming_response, operation, response_callbacks=None):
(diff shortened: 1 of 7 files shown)


bravado/http_future.py
```diff
@@ -101,18 +109,58 @@ def __init__(self, future, response_adapter, operation=None,
...
+
+                swagger_result = incoming_response.swagger_result
+            elif incoming_response.status_code >= 300:
+                raise make_http_exception(response=incoming_response)
+        except exceptions_to_catch as e:
+            if fallback_response:
+                swagger_result = fallback_response(e)
+                incoming_response = None
+            else:
+                raise
```
macisamuele: I would suggest to use ``six.reraise`` for better compatibility

bravado/http_future.py
```diff
@@ -101,18 +109,58 @@ def __init__(self, future, response_adapter, operation=None,
...
+                    self.response_callbacks,
+                )
+
+                swagger_result = incoming_response.swagger_result
+            elif incoming_response.status_code >= 300:
+                raise make_http_exception(response=incoming_response)
+        except exceptions_to_catch as e:
+            if fallback_response:
+                swagger_result = fallback_response(e)
+                incoming_response = None
```
macisamuele: Is there a reason to hide the ``incoming_response`` ?

bravado/http_future.py
```diff
@@ -101,18 +109,58 @@ def __init__(self, future, response_adapter, operation=None,
...
+            else:
+                raise
+
+        return BravadoResponse(
+            result=swagger_result,
+            incoming_response=incoming_response,
+        )
+
+    def result(self, timeout=None):
+        """DEPRECATED: please use the `response()` method instead.
```
macisamuele: Do we really want to deprecate this?
Personally I think that result is useful to get the swagger result object, while the new method (``response``) if useful to get also meta-data, etc.

bravado/http_future.py
```diff
@@ -101,18 +109,58 @@ def __init__(self, future, response_adapter, operation=None,
...
+        :param exceptions_to_catch: Exception classes to catch and call `fallback_response`
+            with. Has no effect if `fallback_response` is not provided. By default, `fallback_response`
+            will be called for read timeout and server errors (HTTP 5XX).
+        :type exceptions_to_catch: List/Tuple of Exception classes.
+        :return: A BravadoResponse instance containing the swagger result and response metadata.
+        """
+        swagger_result = None
+        try:
+            incoming_response = self._get_incoming_response(timeout)
+            if self.operation is not None:
```
macisamuele: Part of this code is duplicated in ``result`` ... could you extract it in an helper method that returns ``incoming_response`` and ``swagger_result``. By doing so ``result`` and ``response`` will use the same method and we'll avoid to maintain two different flows for unmarshaling the HTTP response

bravado/http_future.py
```diff
@@ -13,7 +13,15 @@
 
 from bravado.config_defaults import REQUEST_OPTIONS_DEFAULTS
 from bravado.exception import BravadoTimeoutError
+from bravado.exception import HTTPServerError
 from bravado.exception import make_http_exception
+from bravado.response import BravadoResponse
```
macisamuele: Is this import missing?

bravado/http_future.py
sjaensch: This was to make sure it's always set (an exception can happen during the call to `self._get_incoming_response`). I'm going to refactor it a bit.

bravado/http_future.py
sjaensch: But `response` can be used to get the swagger result object, no? I think it helps in avoiding confusion, and standardizing on one interface. Why would you not want to deprecate it? If/when we actually remove the code is a different question.

macisamuele: agreed

bravado/http_future.py
```diff
@@ -101,26 +109,51 @@ def __init__(self, future, response_adapter, operation=None,
...
+        :type fallback_response: callable that takes an exception and returns a fallback swagger result
+        :param exceptions_to_catch: Exception classes to catch and call `fallback_response`
+            with. Has no effect if `fallback_response` is not provided. By default, `fallback_response`
+            will be called for read timeout and server errors (HTTP 5XX).
+        :type exceptions_to_catch: List/Tuple of Exception classes.
+        :return: A BravadoResponse instance containing the swagger result and response metadata.
+        """
+        incoming_response = None
+        try:
+            swagger_result, incoming_response = self._get_result_and_incoming_response(timeout)
```
macisamuele: I would suggest to split ``incoming_response`` and ``swagger_result`` evaluation.

Let's assume that the service returns HTTP/503 from the request (the exception will be part of ``exceptions_to_catch``).
If you're extracting both objects together you'll not be able to keep trace of ``incoming_response`` as the evaluation of ``swagger_result`` calls ``unmarshal_response`` that will raise the expected bravado exception.
So this method call will return a ``BravadoResponse`` with no information about the incoming response.

My suggestion would be to split them
```python
try:
    incoming_response = self._get_incoming_response(timeout=timeout)
    swagger_result = xxx  # If this raises incoming_response has already be assigned and will be available
    ....
except exceptions_to_catch as e:
    ... 
```

bravado/http_future.py
```diff
@@ -13,7 +13,15 @@
 
 from bravado.config_defaults import REQUEST_OPTIONS_DEFAULTS
 from bravado.exception import BravadoTimeoutError
+from bravado.exception import HTTPServerError
 from bravado.exception import make_http_exception
+from bravado.response import BravadoResponse
+
+
+FALLBACK_EXCEPTIONS = (
```
ArnaudBrousseau: Should we automatically catch exceptions which are mapped to 5xx status codes? Thinking about `bravado.exception.status_map`, here:
https://github.com/Yelp/bravado/blob/cf8ccc26d2f151d67da0f2eab6e0ac0eb9a20eec/bravado/exception.py#L12

bravado/http_future.py
```diff
@@ -101,26 +109,53 @@ def __init__(self, future, response_adapter, operation=None,
...
 
-    @reraise_errors
-    def result(self, timeout=None):
+    def response(self, timeout=None, fallback_response=None, exceptions_to_catch=FALLBACK_EXCEPTIONS):
         """Blocking call to wait for the HTTP response.
 
+        :param timeout: Number of seconds to wait for a response. Defaults to
+            None which means wait indefinitely.
+        :type timeout: float
+        :param fallback_response: callable that accepts an exception as argument and returns the
```
ArnaudBrousseau: This should be renamed to `fallback_result`

bravado/http_future.py
ArnaudBrousseau: I think we should keep `.result` around, otherwise we'll block people from upgrading. Our goal was to stay backwards-compatible no?

bravado/response.py
```diff
@@ -0,0 +1,17 @@
+# -*- coding: utf-8 -*-
+
+
+class BravadoResponse(object):
+    def __init__(self, incoming_response, result):
+        self.incoming_response = incoming_response
+        self.result = result
+        self.response_metadata = BravadoResponseMetadata(
```
ArnaudBrousseau: Let's rename this to "metadata" since we're already in the context of "BravadoResponse"

sjaensch: Sounds good, I've run into this as well. I'll rename it in an upcoming PR.

bravado/http_future.py
sjaensch: Yep we're keeping it, just advising people to use `response()`.

sjaensch:
This PR has been merged into the `9.3.0-release` branch.

PR #362, opened 2018-05-26
Include redirections into HTTP bravado exceptions

macisamuele:
The current PR addresses two aspects of the HTTP exception raised by bravado:
1. handle HTTP/3xx exceptions
currently, in case of redirection errors a generic `HTTPError` is returned, this seems a bit off as the other _families_ of exceptions have a common exception class (ie. `HTTPClientError`)
In this PR I've added the definition of the common `HTTPRedirection` class to handle HTTP/3xx errors and the IANA registered status codes
2. handle the unknown exception of given classes
currently in case of a status code that is not in the list of the standard status codes (recognized by IANA) bravado returns a generic `HTTPError`. This behavior is legit if the status code is not part of the known families of errors (3xx, 4xx and 5xx) but in case of a known family of errors would be better to return the exception class representing the family
An example is in case of 560 status code (random status code not recognized by IANA). With the current setup, bravado returns a generic `HTTPError` while an `HTTPServerError` would be better.

NOTE: the changes introduced by this PRs are backward compatible as:
* we're adding new exception classes that extend the original exceptions
so if users are already catching those exceptions the behavior will be unchanged
* we're modifying the exception returned by unknown status codes
in this case, users could catch those exceptions only by catching `HTTPError`, which again is the base class for the newly raised exceptions

Bonus point: added docstrings for all the HTTP Exception error classes


bravado/exception.py
```diff
@@ -71,10 +71,27 @@ class doesn't exist for a particular HTTP status code, a more
...
+    exc_class = status_map.get(status_code, None)
+
+    if exc_class is None:
+        exception_families = {
+            300: HTTPRedirection,
+            400: HTTPClientError,
+            500: HTTPServerError,
+        }
+        exc_class = exception_families.get(
+            int(status_code / 100) * 100,
```
sjaensch: You could also use the // (integer division) operator.

bravado/exception.py
```diff
@@ -71,10 +71,27 @@ class doesn't exist for a particular HTTP status code, a more
         value of the response.
     :return: An HTTP exception class that can be raised
     """
-    exc_class = status_map.get(response.status_code, HTTPError)
+    status_code = response.status_code
+    exc_class = status_map.get(status_code, None)
```
sjaensch: That `None` is not necessary.

PR #364, opened 2018-05-28
We require at least requests 2.4 for requests.exceptions.ReadTimeout

sjaensch:
This fixes Yelp/bravado-core#271.


andresriancho:
Do we have a new version in pypi?

sjaensch:
Not yet - I don't see a big reason to do a release just for this change. It would require users to bump their bravado version, and if they can do that then they can just bump requests instead (which they have to do either way).

PR #365, opened 2018-05-29
Implement initial BravadoResponseMetadata version

sjaensch:
Besides (obviously) general review feedback, I'd appreciate hearing thoughts about what else we'd want to be present in BravadoResponseMetadata.

### DIFF
--- a/bravado/http_future.py
+++ b/bravado/http_future.py
@@ -131,7 +131,6 @@ class HttpFuture(object):
         use at your own risk.
         """
         incoming_response = None
-        is_fallback_result = False
         exc_info = None
         end_time = None
         try:
@@ -146,14 +145,12 @@ class HttpFuture(object):
             exc_info = sys.exc_info()
             if fallback_result:
                 swagger_result = fallback_result(e)
-                is_fallback_result = True
             else:
                 six.reraise(*exc_info)
 
         response_metadata = BravadoResponseMetadata(
             incoming_response=incoming_response,
             swagger_result=swagger_result,
-            is_fallback_result=is_fallback_result,
             elapsed_time=end_time - self._start_time,
             exc_info=exc_info,
         )
--- a/bravado/http_future.py
+++ b/bravado/http_future.py
@@ -2,6 +2,7 @@
 import sys
 from functools import wraps
 
+import monotonic
 import six
 from bravado_core.content_type import APP_JSON
 from bravado_core.content_type import APP_MSGPACK
@@ -16,6 +17,7 @@ from bravado.exception import BravadoTimeoutError
 from bravado.exception import HTTPServerError
 from bravado.exception import make_http_exception
 from bravado.response import BravadoResponse
+from bravado.response import BravadoResponseMetadata
 
 
 FALLBACK_EXCEPTIONS = (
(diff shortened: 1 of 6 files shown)


setup.py
```diff
@@ -37,6 +37,7 @@
         "pyyaml",
         "requests >= 2",
```
macisamuele: shouldn't this be 2.4 ?

bravado/response.py
```diff
@@ -2,16 +2,47 @@
...
+    NOTE: The `elapsed_time` attribute might be slightly lower than the actual time spent since calling
+    the operation object, as we only start measuring once the call to `HTTPClient.request` returns.
+    Nevertheless, it should be accurate enough for logging and debugging, i.e. determining what went
+    on and how much time was spent waiting for the response.
+
+    WARNING: This interface is considered UNSTABLE. Backwards-incompatible API changes may occur;
+    use at your own risk.
+    """
+
+    def __init__(self, incoming_response, swagger_result, is_fallback_result, elapsed_time, exc_info):
```
macisamuele: I would suggest to not have `is_fallback_result` parameter, but rather having the metadata class to evaluate it.
Saying this because otherwise we'll add potentially a lot of new parameters for information that we want to expose with the metadata.
for example we can evaluate `is_fallback_result` as `exc_info != (None, None, None)`

bravado/http_future.py
```diff
@@ -103,42 +105,61 @@ class HttpFuture(object):
...
                 raise make_http_exception(response=incoming_response)
         except exceptions_to_catch as e:
-            if fallback_response:
-                swagger_result = fallback_response(e)
+            if end_time is None:
+                end_time = monotonic.monotonic()
+            exc_info = sys.exc_info()
+            if fallback_result:
+                swagger_result = fallback_result(e)
+                is_fallback_result = True
```
macisamuele: Check comment below ... but I think that we can avoid having this parameter

bravado/http_future.py
```diff
@@ -103,42 +105,61 @@ class HttpFuture(object):
...
+        WARNING: This interface is considered UNSTABLE. Backwards-incompatible API changes may occur;
+        use at your own risk.
         """
         incoming_response = None
+        is_fallback_result = False
+        exc_info = None
+        end_time = None
         try:
             incoming_response = self._get_incoming_response(timeout)
+            end_time = monotonic.monotonic()
```
macisamuele: ++
I really like this.

bravado/http_future.py
```diff
@@ -103,42 +105,61 @@ class HttpFuture(object):
...
+                end_time = monotonic.monotonic()
+            exc_info = sys.exc_info()
+            if fallback_result:
+                swagger_result = fallback_result(e)
+                is_fallback_result = True
             else:
-                six.reraise(type(e), e, sys.exc_info()[2])
+                six.reraise(*exc_info)
 
+        response_metadata = BravadoResponseMetadata(
```
macisamuele: Nice approach ... this also simplifies the future work to make `BravadoResponseMetadata` class configurable 🥇

bravado/response.py
```diff
@@ -2,16 +2,47 @@
...
+        self.elapsed_time = elapsed_time
+        self.exc_info = exc_info
+
+        # we expose the result to the user through the BravadoResponse object;
+        # we're passing it in to this object in case custom implementations need it
+        self._swagger_result = swagger_result
+
+    @property
+    def status_code(self):
+        return self.incoming_response.status_code
```
macisamuele: `self.incoming_response` is `None` in case of timeout errors .
Do we want to raise an exception in such case?

setup.py
sjaensch: I haven't merged master back into the 9.3 release branch yet.

bravado/response.py
sjaensch: I wasn't sure if these two conditions would always be in sync, but you're right, for now they should be. I've changed it.

bravado/response.py
```diff
@@ -41,8 +40,18 @@ def __init__(self, incoming_response, swagger_result, is_fallback_result, elapse
 
     @property
     def status_code(self):
+        if not self.incoming_response:
+            raise ValueError('No incoming_response present')
```
macisamuele: Feel safer :)
How about having `incoming_response` as a property that raises the exception or having a small helper that asserts the `incoming_response` presence? we can avoid to do the same check all the time

bravado/response.py
```diff
@@ -2,16 +2,56 @@
...
+        self.exc_info = exc_info
+
+        # we expose the result to the user through the BravadoResponse object;
+        # we're passing it in to this object in case custom implementations need it
+        self._swagger_result = swagger_result
+
+    @property
+    def status_code(self):
+        if not self.incoming_response:
+            raise ValueError('No incoming_response present')
```
macisamuele: Could we define an helper method that asserts for this condition? I think that it will simplify testing and will ensure a consistent behaviour (especially if we'll add more properties relying on `incoming_response`)
A possible alternative is to define `incoming_response` as a property (that raises exception)

PR #366, opened 2018-05-31
Implement configurable response metadata class, fallback result killswitch

sjaensch:
I've increased test coverage as well with this branch.

It works, but I think I need to get used to accessing configs like this (`self.operation.swagger_spec.config['bravado']`). On the other hand, passing around a second config object might not be nice either. Thoughts?


bravado/config.py
```diff
@@ -0,0 +1,63 @@
...
+    )
+):
+    @staticmethod
+    def from_config_dict(config):
+        if config is None:
+            config = {}
+        bravado_config = {key: value for key, value in config.iteritems() if key in BravadoConfig._fields}
+        bravado_config = dict(CONFIG_DEFAULTS, **bravado_config)
+        return BravadoConfig(
+            **bravado_config
```
macisamuele: ... does not feel super safe to me
What will happen if we add a new key in `CONFIG_DEFAULTS` and we forgot to trace it into the named tuple?

My suggestion would be to first merge `bravado_config` with `CONFIG_DEFAULTS` and then filter out extraneous keys

bravado/config.py
```diff
@@ -0,0 +1,63 @@
...
+        )
+
+
+def _import_class(fully_qualified_class_str):
```
macisamuele: At the moment this method is used only outside of the module, so either you should refactor a bit the code or you make this function _public_

bravado/http_future.py
```diff
@@ -143,12 +143,13 @@ def response(self, timeout=None, fallback_result=None, exceptions_to_catch=FALLB
...
                 end_time = monotonic.monotonic()
             exc_info = sys.exc_info()
-            if fallback_result:
+            if self.operation and not self.operation.swagger_spec.config['bravado'].disable_fallback_results:
                 swagger_result = fallback_result(e)
             else:
                 six.reraise(*exc_info)
 
-        response_metadata = BravadoResponseMetadata(
+        metadata_class = get_metadata_class(self.operation.swagger_spec.config['bravado'].response_metadata_class)
```
macisamuele: I would refactor a bit the code to have the metadata class saved into the config object.

Doing so allows to have the class type (not string) in the config namedtuple and you don't have to call `get_metadata_class` for each request

bravado/response.py
```diff
@@ -55,3 +60,20 @@ def headers(self):
...
+def get_metadata_class(fully_qualified_class_str):
+    class_to_import = _import_class(fully_qualified_class_str)
+    if not class_to_import:
+        return BravadoResponseMetadata
+
+    if issubclass(class_to_import, BravadoResponseMetadata):
+        return class_to_import
+
+    log.warning(
+        'bravado configuration error: the metadata class {fully_qualified_class_str}\' does not extend '
```
macisamuele: nit: potentially you forgot the opening quote :)
Small note is about logging, please do not format the string yourself but use `%s` and params to logging method (ie. `log.warning('..the metadata class \'%s\' does not extend ...', fully_qualified_class_str)`). In this way the string will be built only if the content has to be logged

bravado/client.py
```diff
@@ -120,15 +120,20 @@ def from_spec(cls, spec_dict, origin_url=None, http_client=None,
         :rtype: :class:`bravado_core.spec.Spec`
         """
         http_client = http_client or RequestsClient()
+        config = config or {}
 
         # Apply bravado config defaults
-        config = dict(CONFIG_DEFAULTS, **(config or {}))
+        bravado_config = BravadoConfig.from_config_dict(config)
+        # remove bravado configs from config dict
+        for key in set(bravado_config).intersection(set(config)):
```
macisamuele: `bravado_config` is a namedtuple object, calling set on it will return the content of the different fields
```python
>>> C = namedtuple('C', 'a, b, c')
>>> instance = C(1,2,3)
>>> print instance
C(a=1, b=2, c=3)
>>> set(instance)
set([1, 2, 3])
>>> set(C._fields)
set(['a', 'c', 'b'])
>>> 
```

bravado/http_future.py
```diff
@@ -143,12 +143,13 @@ def response(self, timeout=None, fallback_result=None, exceptions_to_catch=FALLB
             if end_time is None:
                 end_time = monotonic.monotonic()
             exc_info = sys.exc_info()
-            if fallback_result:
+            if self.operation and not self.operation.swagger_spec.config['bravado'].disable_fallback_results:
```
macisamuele: Seems that you removed the `fallback_result` check 😕

bravado/config.py
sjaensch: Then a test will fail. :)

bravado/config.py
```diff
@@ -0,0 +1,82 @@
...
+        bravado_config = dict(CONFIG_DEFAULTS, **bravado_config)
+        bravado_config['response_metadata_class'] = get_response_metadata_class(
+            bravado_config['response_metadata_class'],
+        )
+        return BravadoConfig(
+            **bravado_config
+        )
+
+
+def get_response_metadata_class(fully_qualified_class_str):
```
macisamuele: Do we want this to be part of the public interface?

PR #367, opened 2018-05-31
Change some metadata names, fix a potential Python 2 memory leak

sjaensch:
Are we ok with these new names?

See https://docs.python.org/2/library/sys.html#sys.exc_info for the other part.

### DIFF
--- a/bravado/http_future.py
+++ b/bravado/http_future.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 import sys
+import traceback
 from functools import wraps
 
 import monotonic
@@ -130,10 +131,7 @@ class HttpFuture(object):
         use at your own risk.
         """
         incoming_response = None
-        # the Python 2 documentation states that we shouldn't assign the traceback to a local variable,
-        # as that would cause a circular reference. We'll use this bool to track whether an exception
-        # was handled instead.
-        caught_exception = False
+        exc_info = None
         end_time = None
         try:
             incoming_response = self._get_incoming_response(timeout)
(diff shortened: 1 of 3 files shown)


bravado/http_future.py
```diff
@@ -141,25 +144,25 @@ def response(self, timeout=None, fallback_result=None, exceptions_to_catch=FALLB
...
-                six.reraise(*exc_info)
+                six.reraise(*sys.exc_info())
 
         metadata_class = self.operation.swagger_spec.config['bravado'].response_metadata_class
         response_metadata = metadata_class(
             incoming_response=incoming_response,
             swagger_result=swagger_result,
             elapsed_time=end_time - self._start_time,
-            exc_info=exc_info,
+            handled_exception_info=sys.exc_info() if caught_exception else None,
```
ArnaudBrousseau: Is that going to work? When I try `sys.exc_info` outside of the `except` block I can't get to it anymore (trying with python 2.7.6):

```
>>> try:
...     raise ValueError('hi')
... except:
...     pass
...
>>> import sys; print(sys.exc_info())
(None, None, None)
>>> try:
...     raise ValueError('hi')
... except:
...     import sys; print(sys.exc_info())
...
(<type 'exceptions.ValueError'>, ValueError('hi',), <traceback object at 0x7f84248d23f8>)
```

sjaensch: Yeah, looks like it doesn't work. :( Interestingly, on Travis the test only fails on Python 3... not sure why.
I've rewritten it so that we'll have a string representation of the traceback, as we only need it for logging anyway.

bravado/response.py
```diff
@@ -29,10 +29,18 @@ class BravadoResponseMetadata(object):
...
     """
 
-    def __init__(self, incoming_response, swagger_result, elapsed_time, exc_info):
+    def __init__(self, incoming_response, swagger_result, elapsed_time, handled_exception_info):
+        """
+        :param incoming_response: a subclass of bravado_core.response.IncomingResponse.
+        :param swagger_result: the unmarshalled result that is being returned to the user.
+        :param elapsed_time: float containing an approximate elapsed time since creating the future, in seconds.
+        :param handled_exception_info: sys.exc_info() data if an exception was caught and handled as
+            part of a fallback response; note that the third element in the list is a string representation
```
macisamuele: I would say that the third element should contain the string representation of the traceback and not the traceback due to circular reference.

It feels strange to me saying that the third element is the string representation as it the caller to decide what it is

sjaensch: Heh, but the caller will be bravado code. I see this more as a documentation for a) people wanting to use the metadata and b) people wanting to implement their own version of the class. So we know what the argument will be. :)

PR #369, opened 2018-06-05
Validate and unmarshal the fallback result

sjaensch:
This changes the expected fallback result value to be a Python representation of the wire data, which I think is more intuitive to use. The change also causes response validation to run for the fallback result data, which could help in catching errors.

On the other hand, it potentially makes testing this case harder, as we can't just mock BravadoResponse.result to return whatever the fallback_result callback returns - unmarshalling needs to occur for it to be correct. Thoughts?

### DIFF
--- a/bravado/http_future.py
+++ b/bravado/http_future.py
@@ -18,6 +18,7 @@ from bravado.exception import BravadoTimeoutError
 from bravado.exception import HTTPServerError
 from bravado.exception import make_http_exception
 from bravado.response import BravadoResponse
+from bravado.response import FallbackIncomingResponse
 
 
 FALLBACK_EXCEPTIONS = (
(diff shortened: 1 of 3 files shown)


ArnaudBrousseau:
Because fallback results are specified in code they should ideally be specified in terms of what the code is most used to deal with: models and unmarshalled values.

If we agree on this, then we need a `marshall` step to take the value returned by the `fallback_result` callback and transform it to a wire format. Feels a bit odd?

I agree with the need to validate results as they're unmarshalled. That'll prevent a class of bugs (invalid fallback results from the swagger spec POV)

sjaensch:
We're going to go with an approach where `fallback_result` returns an unmarshalled value. We'll document how to get and use the model classes.

PR #372, opened 2018-06-12
Introduce RequestConfig and a way to force returning fallback results

sjaensch:
First iteration at supporting forcing fallback results. Since that would require a backwards-incompatible change, I went and tried to rework it in a way that would make it possible to do these things without breaking compatibility in the future.

### DIFF
--- a/bravado/client.py
+++ b/bravado/client.py
@@ -54,7 +54,7 @@ from six import iteritems
 from six import itervalues
 
 from bravado.config import BravadoConfig
-from bravado.config import REQUEST_OPTIONS_DEFAULTS
+from bravado.config import RequestConfig
 from bravado.docstring_property import docstring_property
 from bravado.requests_client import RequestsClient
 from bravado.swagger_model import Loader
@@ -239,27 +239,20 @@ class CallableOperation(object):
         log.debug(u'%s(%s)', self.operation.operation_id, op_kwargs)
         warn_for_deprecated_op(self.operation)
 
-        # Apply request_options defaults
-        request_options = dict(
-            REQUEST_OPTIONS_DEFAULTS,
-            **(op_kwargs.pop('_request_options', {})))
+        # Get per-request config
+        request_options = op_kwargs.pop('_request_options', {})
+        request_config = RequestConfig.from_request_options_dict(request_options, self.also_return_response)
 
         request_params = construct_request(
             self.operation, request_options, **op_kwargs)
 
         http_client = self.operation.swagger_spec.http_client
 
-        # Per-request config overrides client wide config
-        also_return_response = request_options.get(
-            'also_return_response',
-            self.also_return_response,
-        )
-
         return http_client.request(
             request_params,
             operation=self.operation,
-            response_callbacks=request_options['response_callbacks'],
-            also_return_response=also_return_response)
+            request_config=request_config,
+        )
 
 
 def construct_request(operation, request_options, **op_kwargs):
--- a/bravado/client.py
+++ b/bravado/client.py
@@ -241,7 +241,7 @@ class CallableOperation(object):
 
         # Get per-request config
         request_options = op_kwargs.pop('_request_options', {})
-        request_config = RequestConfig.from_request_options_dict(request_options, self.also_return_response)
+        request_config = RequestConfig(request_options, self.also_return_response)
 
         request_params = construct_request(
             self.operation, request_options, **op_kwargs)
--- a/bravado/config.py
+++ b/bravado/config.py
@@ -3,6 +3,10 @@ import logging
 from collections import namedtuple
 from importlib import import_module
 
+import typing  # noqa: F401
+from bravado_core.operation import Operation  # noqa: F401
+from bravado_core.response import IncomingResponse  # noqa: F401
+
 from bravado.response import BravadoResponseMetadata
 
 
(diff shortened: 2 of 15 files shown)


bravado/client.py
```diff
@@ -239,27 +239,20 @@ def __call__(self, **op_kwargs):
         log.debug(u'%s(%s)', self.operation.operation_id, op_kwargs)
         warn_for_deprecated_op(self.operation)
 
-        # Apply request_options defaults
-        request_options = dict(
-            REQUEST_OPTIONS_DEFAULTS,
-            **(op_kwargs.pop('_request_options', {})))
+        # Get per-request config
+        request_options = op_kwargs.pop('_request_options', {})
+        request_config = RequestConfig.from_request_options_dict(request_options, self.also_return_response)
```
macisamuele: I have some doubts about setting a named tuple to keep request config data.
At the moment bravado internally needs/uses few keys so that seems reasonable, but there are few keys (ie. `tcp_nodelay` used by fido) that are actually defined as HttpClient specific.
Using a namedtuple will prevent exchange of request options that are strictly dependent on HTTP client and/or we'll have to expand the namedtuple attributes "everytime" it could happen.

Even if that was not an intended feature, I think that it would be cool to keep.
What about converting the named tuple that keeps argument in input and allows and stores kwargs too?
In this way to could have _stricter_ control on the request configuration passed around and give free space to any further extension

As this change have the potential to reduce extendibility I would like to iterate a bit more on this.

bravado/config.py
```diff
@@ -52,6 +53,33 @@ def from_config_dict(config):
...
+            'headers',
+            'use_msgpack',
+            'timeout',
+        ]
+    )
+):
+    @staticmethod
+    def from_request_options_dict(request_options, also_return_response_default):
+        request_config = {
+            key: request_options.get(key, REQUEST_OPTIONS_DEFAULTS.get(key))
```
macisamuele: As we need to do a backward incompatible change and we're renaming the concept of request_options into request_config do you mind renaming the constant accordingly?

bravado/http_future.py
```diff
@@ -136,9 +135,25 @@ def response(self, timeout=None, fallback_result=None, exceptions_to_catch=FALLB
...
+                if self.operation.swagger_spec.config['bravado'].disable_fallback_results:
+                    log.warning(
+                        'force_fallback_result set in request options and disable_fallback_results '
+                        'set in client config; not using fallback result.'
+                    )
+                else:
+                    # raise an exception to trigger fallback result handling
+                    fake_exception_type = exceptions_to_catch[0]
+                    args = (incoming_response,) if issubclass(fake_exception_type, HTTPError) else ()
+                    raise fake_exception_type(*args)
```
macisamuele: I'm wondering it this will always be possible and I think that it is not.

At the moment you're customising the value of `args` in case of an `HTTPError`; what happen if the first exception in `exceptions_to_catch` needs additional arguments to be instantiated?
An example could be [`aiohttp.client_exceptions.ClientResponseError`](https://github.com/aio-libs/aiohttp/blob/965ddc1ea3a8fbf1f8d0f64efa0f117a8ebb64ab/aiohttp/client_exceptions.py#L43)

Other than potential issues in creating the exception to raise this could have negative effects over the monitoring systems that we could set as we'll not be able to distinguish if the exception was a "real" exception or an injected one.

To workaroud those issues I would suggest to define a custom exception (ie. `DegradedResponseError`), raise it and ensure that only that exception will be catched.
This allows you to:
 * not have to care about the initialisation of the exception instance (as it will be under complete bravado control)
 * be able to discriminate on the logs what exception have been raised (so monitoring systems will not alert the teams in case of a flood of requests which forces force_degradation)

sjaensch: @macisamuele good points, and as we've discussed I'm not sure about what the best approach overall is. Earlier on I used a BravadoTimeoutError; the problem with not using an exception from `exceptions_to_catch` is that the `fallback_result` function is not guaranteed to be able to handle that type of exception.

This PR gives users some control over which exception is going to be thrown in the `force` case, although I'll admit it's quite unintuitive and doesn't cover all cases.

If we're ok with introducing this new custom exception type then I'll go ahead and change this PR accordingly - users can then adapt their callback functions to treat it any way they want to.

bravado/client.py
sjaensch: I agree it could prove to be problematic to be that strict; let's discuss offline. One thing I'll mention is that `tcp_nodelay` is _not_ passed through as a request_option, so is unrelated to this change. This is why the test for it continues to pass. ;)

bravado/http_future.py
```diff
@@ -132,6 +133,9 @@ def response(self, timeout=None, fallback_result=None, exceptions_to_catch=FALLB
         incoming_response = None
         exc_info = None
         request_end_time = None
+        if self.request_config.force_fallback_result:
+            exceptions_to_catch = tuple(chain(exceptions_to_catch, (ForcedFallbackResultError,)))
```
macisamuele: Is this needed? Could we just override `exceptions_to_catch` with `ForcedFallbackResultError`

sjaensch: Like discussed offline: Let's stick to the approach of only throwing the `ForcedFallbackResultError` in case no other exception is thrown, as that makes it significantly easier to handle cleanly for async HTTP clients. I've moved it to the end of the block.

macisamuele:
Feel free to merge this

Issue #373, opened 2018-06-14
Should not rewrite custom auth headers from _request_options 


aviskase:
Hi, I'm using bravado for end-to-end checks. One use case is : 
1. Use `default_apikey` for every request
2. Create `new_apikey` using API
3. Check that requests with `new_apikey` applied work.

Currently to make this work one should wrap SwaggerClient/RequestsClient or create separate SwaggerClients for each api key. Reason is that `ApiKeyAuthenticator.apply` overwrites any custom auth headers from `_request_options`. 

It would be better if  `ApiKeyAuthenticator.apply` could  skip setting `request.headers[self.param_name]` if this header already exists.

I suppose similar thing is applicable to other Authenticator variations.

PR #374, opened 2018-06-16
Support per-request API key header overwrite by _request_options

aviskase:
Fixes problem from #373 . Probably there might be some more uses cases (not only a header).


sjaensch:
@aviskase thanks for your contribution! I've just released bravado 10.1.0 which contains your changes.

PR #375, opened 2018-06-18
Introduce testing mocks for responses

sjaensch:
Two notable changes from our internal implementation:

1. The \_\_call__ signatures are different. Internally we have `fallback_result` as the first argument (and no `exceptions_to_catch`). I've added a test to make sure these stay in sync with the signature of HttpFuture.response.
2. People are now expected to pass in an exception instance to DegradedBravadoResponseMock. This is necessary so they can test HTTPServerErrors, HTTP client errors and so forth, as these exceptions might require arguments to be provided for instantiation.
3. `metadata._swagger_result` was not set correctly in the degraded response case; this is fixed and tested.

### DIFF
--- a/bravado/testing/response_mocks.py
+++ b/bravado/testing/response_mocks.py
@@ -1,9 +1,17 @@
 # -*- coding: utf-8 -*-
-import mock
+from bravado_core.response import IncomingResponse
 
 from bravado.exception import BravadoTimeoutError
 from bravado.http_future import FALLBACK_EXCEPTIONS
 from bravado.response import BravadoResponseMetadata
+
+
+class FakeIncomingResponse(IncomingResponse):
+    def __init__(self, status_code, **kwargs):
+        self.headers = {}
+        self.status_code = status_code
+        for name, value in kwargs.items():
+            setattr(self, name, value)
 
 
 class BravadoResponseMock(object):
@@ -17,10 +25,7 @@ class BravadoResponseMock(object):
             self._metadata = metadata
         else:
             self._metadata = BravadoResponseMetadata(
-                incoming_response=mock.Mock(
-                    status_code=200,
-                    headers={},
-                ),
+                incoming_response=FakeIncomingResponse(status_code=200),
                 swagger_result=self._result,
                 start_time=1528733800,
                 request_end_time=1528733801,
@@ -61,6 +66,8 @@ class FallbackResultBravadoResponseMock(object):
             )
 
     def __call__(self, timeout=None, fallback_result=None, exceptions_to_catch=FALLBACK_EXCEPTIONS):
+        assert callable(fallback_result), 'You\'re using FallbackResultBravadoResponseMock without a callable ' + \
+            'fallback_result. Either provide a callable or use BravadoResponseMock.'
         self._fallback_result = fallback_result(self._exception)
         self._metadata._swagger_result = self._fallback_result
         return self
--- a/bravado/testing/response_mocks.py
+++ b/bravado/testing/response_mocks.py
@@ -1,0 +1,58 @@
+# -*- coding: utf-8 -*-
+from bravado.exception import BravadoTimeoutError
+from bravado.http_future import FALLBACK_EXCEPTIONS
+from bravado.response import BravadoResponseMetadata
+
+
+class BravadoResponseMock(object):
+    def __init__(self, result):
+        self._result = result
+
+    def __call__(self, timeout=None, fallback_result=None, exceptions_to_catch=FALLBACK_EXCEPTIONS):
+        return self
+
+    @property
+    def result(self):
+        return self._result
+
+    @property
+    def metadata(self):
+        return BravadoResponseMetadata(
+            incoming_response='incoming response',
+            swagger_result=self._result,
+            start_time=1528733800,
+            request_end_time=1528733801,
+            handled_exception_info=None,
+            request_config=None,
+        )
+
+
+class DegradedBravadoResponseMock(object):
+    def __init__(self, exception=BravadoTimeoutError()):
+        self._exception = exception
+
+    def __call__(self, timeout=None, fallback_result=None, exceptions_to_catch=FALLBACK_EXCEPTIONS):
+        self._fallback_result = fallback_result(self._exception)
+        return self
+
+    @property
+    def result(self):
+        return self._fallback_result
+
+    @property
+    def metadata(self):
+        return BravadoResponseMetadata(
+            incoming_response=None,
+            swagger_result=self._fallback_result,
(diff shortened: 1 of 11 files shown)


bravado/testing/response_mocks.py
```diff
@@ -0,0 +1,58 @@
...
+    def result(self):
+        return self._result
+
+    @property
+    def metadata(self):
+        return BravadoResponseMetadata(
+            incoming_response='incoming response',
+            swagger_result=self._result,
+            start_time=1528733800,
+            request_end_time=1528733801,
```
ArnaudBrousseau: Same for `request_end_time`. People might want the ability to override this.

bravado/testing/response_mocks.py
```diff
@@ -0,0 +1,58 @@
...
+            handled_exception_info=None,
+            request_config=None,
+        )
+
+
+class DegradedBravadoResponseMock(object):
```
ArnaudBrousseau: Add a docstring to explain how this class is supposed to be used, potentially with an example.

tests/testing/response_mocks_test.py
```diff
@@ -0,0 +1,62 @@
...
+from bravado.testing.response_mocks import DegradedBravadoResponseMock
+from bravado.testing.response_mocks import make_bravado_response
+
+
+@pytest.fixture
+def mock_result():
+    return mock.Mock(name='mock result')
+
+
+def test_response_mock_signatures():
```
ArnaudBrousseau: ++

bravado/testing/response_mocks.py
```diff
@@ -0,0 +1,58 @@
...
+            incoming_response=None,
+            swagger_result=self._fallback_result,
+            start_time=1528733800,
+            request_end_time=1528733801,
+            handled_exception_info=(self._exception.__class__, self._exception, 'Traceback'),
+            request_config=None,
+        )
+
+
+def make_bravado_response(result, degraded=False):
```
ArnaudBrousseau: I actually don't like this function much, because `result` is only used when the response is not degraded. Let's remove it? It's not that much hassle to write `BravadoResponseMock(result)` and `DegradedBravadoResponseMock()` in tests (as opposed to `make_bravado_response(result)` and `make_bravado_response(result=None, degraded=False)`, resp.)

bravado/testing/response_mocks.py
```diff
@@ -0,0 +1,58 @@
...
+    @property
+    def result(self):
+        return self._result
+
+    @property
+    def metadata(self):
+        return BravadoResponseMetadata(
+            incoming_response='incoming response',
+            swagger_result=self._result,
+            start_time=1528733800,
```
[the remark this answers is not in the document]
macisamuele: I think that we could use the same strategy as for `result`. Defining the default metadata object into `BravadoResponseMock` gives the flexibility to change it if needed without adding too much complexity on testing classes or their usage

bravado/testing/response_mocks.py
macisamuele: I do actually think that `DegradedBravadoResponseMock` should use `result` as tester would like to check that the fallback result is used (and it will be set into `BravadoResponse.result`)

bravado/testing/response_mocks.py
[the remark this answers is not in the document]
sjaensch: I've made it configurable; let me know what you think.

bravado/testing/response_mocks.py
sjaensch: I went a bit further and wrote documentation. :)

bravado/testing/response_mocks.py
sjaensch: I'm not sure what you meant with that @macisamuele. I've removed the function as I agree it doesn't really provide much value; I've also renamed ``DegradedBravadoResponseMock`` so it's consistent with other names in bravado.

bravado/testing/response_mocks.py
```diff
@@ -0,0 +1,74 @@
...
+                incoming_response=None,
+                swagger_result=None,  # we're going to set it later
+                start_time=1528733800,
+                request_end_time=1528733801,
+                handled_exception_info=(self._exception.__class__, self._exception, 'Traceback'),
+                request_config=None,
+            )
+
+    def __call__(self, timeout=None, fallback_result=None, exceptions_to_catch=FALLBACK_EXCEPTIONS):
+        self._fallback_result = fallback_result(self._exception)
```
macisamuele: What will happen if `fallback_result` param is not passed?
Could make sense to define a default handling method?

bravado/testing/response_mocks.py
```diff
@@ -0,0 +1,74 @@
...
+    Please check the documentation for further information.
+    """
+
+    def __init__(self, result, metadata=None):
+        self._result = result
+        if metadata:
+            self._metadata = metadata
+        else:
+            self._metadata = BravadoResponseMetadata(
+                incoming_response=mock.Mock(
```
macisamuele: I'm wondering if we can do something different to have a "mock" response without having mock library.
An example could be found on https://github.com/Yelp/bravado-core/blob/master/tests/profiling/unmarshal_response_profiler_test.py#L6

Doing this will prevent us to have the testing extra dependency.

In general that is not a problem, but I don't think will be un-realistic having `bravado[testing]==X.X.X` in the production requirements and it will add `mock` dependency in production.

bravado/testing/response_mocks.py
sjaensch: Not passing it would mean there's either a bug in the code being tested or the wrong mock is being used; I'd prefer it to fail. I've added an assertion to make it more descriptive.

bravado/testing/response_mocks.py
sjaensch: Good suggestion! I'm working that in.

bravado/testing/response_mocks.py
```diff
@@ -0,0 +1,81 @@
+# -*- coding: utf-8 -*-
+from bravado_core.response import IncomingResponse
+
+from bravado.exception import BravadoTimeoutError
+from bravado.http_future import FALLBACK_EXCEPTIONS
+from bravado.response import BravadoResponseMetadata
+
+
+class FakeIncomingResponse(IncomingResponse):
```
macisamuele: Small naming suggestion: could we use `IncomingResponseMock` like for the other testing classes?
From documentation side I think that would be super helpful to have an example of usage

sjaensch: I'm not sure yet how users would take advantage of `IncomingResponseMock`, i.e. what the testing scenarios would be. I'll add something once that's more clear. :)

PR #376, opened 2018-06-22
Support non-callable fallback results, stabilize API

sjaensch:
So I wanted to try out how it would look like if we supported non-callable fallback results, and it's really a very small change. It seems nicer to use and less error prone in the common case.

Do we expect the `callable(fallback_result)` to be enough to reliably identify the two cases? I wasn't able to come up with an example where that wouldn't be the case, but I'd like to hear your input.

Also, do you feel we're in a place where we can commit to these API? Or should we wait with removing the warnings?

### DIFF
--- a/bravado/http_future.py
+++ b/bravado/http_future.py
@@ -123,8 +123,7 @@ class HttpFuture(object):
         :type timeout: float
         :param fallback_result: either the swagger result or a callable that accepts an exception as argument
             and returns the swagger result to use in case of errors
-        :type fallback_result: either any type of object that represents a valid Swagger result or a callable that
-            takes an exception and returns a fallback swagger result
+        :type fallback_result: Optional[Union[Any, Callable[[Exception], Any]]]
         :param exceptions_to_catch: Exception classes to catch and call `fallback_result`
             with. Has no effect if `fallback_result` is not provided. By default, `fallback_result`
             will be called for read timeout and server errors (HTTP 5XX).
--- a/bravado/http_future.py
+++ b/bravado/http_future.py
@@ -32,6 +32,9 @@ FALLBACK_EXCEPTIONS = (
 )
 
 
+SENTINEL = object()
+
+
 class FutureAdapter(object):
     """
     Mimics a :class:`concurrent.futures.Future` regardless of which client is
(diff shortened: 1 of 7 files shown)


bravado/http_future.py
```diff
@@ -112,23 +115,21 @@ def __init__(self, future, response_adapter, operation=None,
...
 
         :param timeout: Number of seconds to wait for a response. Defaults to
             None which means wait indefinitely.
         :type timeout: float
-        :param fallback_result: callable that accepts an exception as argument and returns the
-            swagger result to use in case of errors
-        :type fallback_result: callable that takes an exception and returns a fallback swagger result
+        :param fallback_result: either the swagger result or a callable that accepts an exception as argument
+            and returns the swagger result to use in case of errors
+        :type fallback_result: either any type of object that represents a valid Swagger result or a callable that
```
macisamuele: This seems to be the same as in param, could we define the type in a more accurate way?
maybe with typing annotation style?

bravado/http_future.py
```diff
@@ -165,10 +166,11 @@ def response(self, timeout=None, fallback_result=None, exceptions_to_catch=FALLB
             exc_info = list(sys.exc_info()[:2])
             exc_info.append(traceback.format_exc())
             if (
-                fallback_result and self.operation
+                fallback_result is not SENTINEL
+                and self.operation
                 and not self.operation.swagger_spec.config['bravado'].disable_fallback_results
             ):
-                swagger_result = fallback_result(e)
+                swagger_result = fallback_result(e) if callable(fallback_result) else fallback_result
```
macisamuele: To me `callable` in general looks a good fit, as bravado-core models or primitives (ie.` int`, `bool`, etc) are not callable.

I do think that the approach could still lead to weird behaviours in case of `format` is used.
The scenario that could lead to issues could be if the bravado client has a custom format that returns callable objects
```python
def f_raise():
    raise ValueError()
def foo(*args, **kwargs): return 1
def bar(*args, **kwargs): return 2
_FUNCTIONS = {f.__name: f for f in [foo, bar]}

format = SwaggerFormat(
    format='function-format',
    to_wire=lambda func: func.__name__,
    to_python=lambda func_string: _FUNCTIONS.get(func_string, f_raise),
    validate=lambda *args, **kwargs: None
)
 ```

This is a case that _theoretically_ is supported even if I don't know if currently users are using it and/or why should someone do something like that.

So the question is: do we want to support such case?
If yes then probably `callable` is not a good fit and most probably we have no real way to determine if `fallback_result` is the function to call in case of exception or the result that should be returned
If no then `callable` is a good fit (eventually documentation should state that issue).

File: docs/source/advanced.rst [hunk 3]
```diff
@@ -169,25 +169,30 @@ By default, if the server returns an error or doesn't respond in time, you have
...
 
+Handling error types differently
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Sometimes, you might want to treat timeout errors differently from server errors. To do this you may
+pass in a callable as ``fallback_result`` argument. The callable takes one mandatory argument: the exception
+that would have been raised normally. This allows you to return different results based on the type of error
+(e.g. a :class:`.BravadoTimeoutError`) or, if a server response was received, on any data pertaining
+to that response, like the HTTP status code. Subclasses of :class:`.HTTPError` have a ``response`` attribute
+that provides access to that data.
```
macisamuele: Could you add a call example?

bravado/http_future.py
sjaensch: Can you provide an actual example of what such a format would do? Otherwise I wouldn't worry about it.

macisamuele: An example is available on https://gist.github.com/macisamuele/846c27c44569e1e06b513a82ee9f8070

We discussed, in person, about it and we agreed that such use case seems mostly a "misuse" of 
`SwaggerFormat` and so we should not be worried about it.

PR #377, opened 2018-06-25
Add connection error support

macisamuele:
The main objective of this PR is to provide support to `ConnectionError`s as is currently provided for `TimeoutError`s.
This aims to address also @bplotnick suggestions in PR #346.

Details of this PR code changes

1. Integration tests base class is no longer `TestServerRequestsClient`. This is intended done in order to move the base integration tests suite into `bravado.testing` module so external http clients (ie. bravado-asyncio) could be able to run the same tests that we run for the http clients provided by bravado.
NOTE: moving integration tests into `bravado.testing` is not addressed here,  a new PR will be provided in the future
2. changed the way that exceptions are raised:
 * Timeout classes are not saved as future attribute. This is done as futures are created for every request the timeout class was being created _every time_ anyway and because it could be possible that classes are not _compatible_ each others (ie. class 1 and class 2 requires different arguments to be allocated and the new class extends both of them). This was not an issue until now as `timeout_errors` contained a single item, but with connection errors (at least with fido client) we have more that one exception class
 * Re-raising an exception uses an hack to make sure that we'll be able to instantiate the error class without changing the properties (attributes and methods) that the original exception class provides

### DIFF
--- a/bravado/http_future.py
+++ b/bravado/http_future.py
@@ -43,7 +43,22 @@ class FutureAdapter(object):
     """
 
     # Make sure to define the timeout errors associated with your http client
-    timeout_errors = []
+    timeout_errors = ()
+
+    def _raise_error(self, base_exception_class, class_name_suffix, exception):
+        error_type = type(
+            '{}{}'.format(self.__class__.__name__, class_name_suffix),
+            (exception.__class__, base_exception_class),
+            dict(),
+        )
+        six.reraise(
+            error_type,
+            error_type(*exception.args),
+            sys.exc_info()[2],
+        )
+
+    def _raise_timeout_error(self, exception):
+        self._raise_error(BravadoTimeoutError, 'Timeout', exception)
 
     def result(self, timeout=None):
         """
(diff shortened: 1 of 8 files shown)


bravado/http_future.py
```diff
@@ -58,32 +83,18 @@ def result(self, timeout=None):
...
-            setattr(
-                self.future, '__timeout_error_type',
-                type(
-                    '{}Timeout'.format(self.future.__class__.__name__),
-                    tuple(list(timeout_errors) + [BravadoTimeoutError]),
-                    dict(),
-                ),
-            )
+        timeout_errors = tuple(self.future.timeout_errors or ())
+        connection_errors = tuple(self.future.connection_errors or ())
```
sjaensch: Could we make this code backwards compatible? I think something like

    connection_errors = getattr(self.future, 'connection_errors', ())

I don't think it's a problem to use whatever the value of that attribute is if it is defined.

macisamuele: Sure ... keeping backward compatibility is something to aim for :)
I'm not against using `getattr` but I would understand how the change could provide backward incompatible changes (so next time I'll think to it too)

macisamuele: We talked about it and well the change is backward compatible as `self.future` in an instance of `bravado.http_future. FutureAdapter` which defines those attributes.
So even if http clients are not defining them the attribute will be available with an empty tuple as default value

PR #378, opened 2018-06-25
Extract integration testing tools into bravado.testing module

macisamuele:
The idea of this PR is to expose into `bravado.testing` module the base integration test suite used for http clients.

My expectation is that exposing this base class will enhance testing of external defined http clients (ie. https://github.com/sjaensch/bravado-asyncio) as they will have exposure to the same tools used to test the internal http clients.

NOTE: all the fixtures defined in `bravado/integration/conftest.py` are now integrated in `IntegrationTestingServicesAndClient`, so all test classes extending `IntegrationTestsBaseClass` will have them without having to play around pytest configurations.


bravado/testing/integration_test.py
```diff
@@ -1,54 +1,256 @@
...
-    """
-    Generic class to run integration tests with the different HTTP clients definitions
-    """
 
-    http_client_type = None
-    http_future_adapter_type = None
-    connection_errors_exceptions = None
+@bottle.get('/swagger.json')
+def swagger_spec():
+    return SWAGGER_SPEC_DICT
```
sjaensch: As discussed offline: if bottle doesn't set the content type automatically then let's add it here too.

bravado/testing/integration_test.py
```diff
@@ -1,54 +1,256 @@
...
+@bottle.post('/double')
+def double():
+    x = bottle.request.params['number']
+    return str(int(x) * 2)
+
+
+@bottle.get('/echo')
+def echo():
+    bottle.response.content_type = APP_JSON
+    return {'message': bottle.request.params['message']}
```
sjaensch: Let's set the content type here as well.

bravado/testing/integration_test.py
macisamuele: Actually setting content_type to application/json is not needed as bottle does that by default.
I'm going to remove them

macisamuele:
Rebased on latest master and removed usage of standard port for `not_answering_http_server` to fix internal build failure.

bravado/testing/integration_test.py
```diff
@@ -245,7 +245,7 @@ def wait_unit_service_starts(url, max_wait_time=10):
     @pytest.fixture(scope='session')
     def not_answering_http_server(self):
         # Nobody could listen on such address, so TCP 3 way handshake will fail
-        yield 'http://0.0.0.0'
+        yield 'http://0.0.0.0:8080'
```
sjaensch: This is still tricky, as that port could be in use as well. We typically use the ephemeral-port-reserve package for such cases.

Is this case really worth doing an "integration test" for? Especially since in the end you're using mock to set the side effect anyway; seems like doing a more unit-testy approach would work better (i.e. no actual TCP connection ever attempted).

sjaensch: Also, just for reference, listening on 0.0.0.0 typically means "listen on all interfaces". So you can very much listen on that address, you just can't use it as a target when connecting.

macisamuele: There are two tests that are using that fixture
 * `test_connection_errors_are_thrown_as_BravadoConnectionError` which sends a real http request (and TCP connection fails)
 * `test_connection_errors_are_catchable_with_original_exception_types` which sends a request, but the the future is mocked out so it does not really matter

About the IP address, you're right that `0.0.0.0` means _listen on all interfaces_  and apparently this gets routed to `127.0.0.1`.
```
maci@dev31-uswest1adevc:~$ curl -v 0.0.0.0
* Rebuilt URL to: 0.0.0.0/
* Hostname was NOT found in DNS cache
*   Trying 0.0.0.0...
* Connected to 0.0.0.0 (127.0.0.1) port 80 (#0)
> GET / HTTP/1.1
> User-Agent: curl/7.35.0
> Host: 0.0.0.0
> Accept: */*
```

The main issue in using `ephemeral-port-reserve` is that in order to reserve a port the package accepts requests on it and this makes complicated to have a refused TCP connection in a reliable way.

I'm going to change the way this fixture is implemented to ensure that a "test handled" port is used so we don't get influenced by the running environment

PR #379, opened 2018-06-29
Generalise integration test for non string headers

macisamuele:
The main goal of this PR is to generalise the existing `test_boolean_header` integration test to check for non-string headers.

In order to guarantee a stable generation of the HTTP request I made sure that the test communicates with an endpoint ``GET /headers`` which returns type and representation of all the headers starting with a defined prefix (to simplify assertions)


bravado/testing/integration_test.py
```diff
@@ -412,15 +428,34 @@ def test_post_request(self, swagger_http_server):
...
 
-        assert response.text == self.encode_expected_response(ROUTE_1_RESPONSE)
+        expected_header_representations = {
+            header_name: repr(
+                str(header_value.decode())
+                if isinstance(header_value, bytes)
+                else str(header_value)
+            )
+            for header_name, header_value in headers.items()
+        }
```
sjaensch: Please provide the actual expected headers; this will make the test more obvious to read and to understand, and it provides us (humans) with the ability to make sure things really are the way we want them to be. :)

PR #391, opened 2018-10-12
Use CSafeLoader if it is available

bplotnick:
This tracks the internal ticket COREBACK-6978. It is a pretty common idiom we use already in other places to use the c-based loader (which is much faster) if it is available.

### DIFF
--- a/bravado/swagger_model.py
+++ b/bravado/swagger_model.py
@@ -5,6 +5,10 @@ import os
 import os.path
 
 import yaml
+try:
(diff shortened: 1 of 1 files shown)


bravado/swagger_model.py
```diff
@@ -5,6 +5,10 @@
 import os.path
 
 import yaml
+try:
```
macisamuele: nit. I'm wondering if it would be better to move this into `compact.py` (as we do for `json` vs `simplejson`)
Anyway this is not blocking.

PR #399, opened 2018-12-05
Add option to create new session per request in RequestsClient

hoou:


bravado/requests_client.py
```diff
@@ -151,7 +154,7 @@ def request(self, request_params, operation=None, request_config=None):
         sanitized_params, misc_options = self.separate_params(request_params)
 
         requests_future = RequestsFutureAdapter(
-            self.session,
+            copy.deepcopy(self.session) if self.session_per_request else self.session,
```
macisamuele: I'm think that calling `requests.Session()` instead of `copy.deepcopy(self.session)` would be clearer.
Is there a specific reason to use the copy instead of the creation of a new object?

PR #402, opened 2018-12-19
Pass correct default value for proxies to requests function

### LINKED ISSUE #401: AttributeError: 'NoneType' object has no attribute 'setdefault' if HTTP_PROXY is set
Okay, this was a nice bit of detective work. I'm calling SwaggerClient.from_url(...), which gives me
```
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../../virtualenv/python3.6.3/lib/python3.6/site-packages/bravado/client.py:96: in from_url
    spec_dict = loader.load_spec(spec_url)
../../../virtualenv/python3.6.3/lib/python3.6/site-packages/bravado/swagger_model.py:104: in load_spec
    self.request_headers,
../../../virtualenv/python3.6.3/lib/python3.6/site-packages/bravado/http_future.py:213: in result
    incoming_response = self._get_incoming_response(timeout)
../../../virtualenv/python3.6.3/lib/python3.6/site-packages/bravado/http_future.py:97: in wrapper
    return func(self, *args, **kwargs)
../../../virtualenv/python3.6.3/lib/python3.6/site-packages/bravado/http_future.py:228: in _get_incoming_response
    inner_response = self.future.result(timeout=timeout)
../../../virtualenv/python3.6.3/lib/python3.6/site-packages/bravado/requests_client.py:310: in result
    self.misc_options['ssl_cert'],
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
self = <requests.sessions.Session object at 0x2af3ea842390>
url = 'http://localhost:29593/swagger.json', proxies = None, stream = None
verify = True, cert = None
    def merge_environment_settings(self, url, proxies, stream, verify, cert):
        """
            Check the environment and merge it with some settings.
    
            :rtype: dict
            """
        # Gather clues from the surrounding environment.
        if self.trust_env:
            # Set environment's proxies.
            no_proxy = proxies.get('no_proxy') if proxies is not None else None
            env_proxies = get_environ_proxies(url, no_proxy=no_proxy)
            for (k, v) in env_proxies.items():
>               proxies.setdefault(k, v)
E               AttributeError: 'NoneType' object has no attribute 'setdefault'
```
This seems to be caused by a bug in Bravado. In particular, in requests_client.py on line 305, requests.Session.merge_environment_settings is called with None for the `proxies` (second) parameter, which will work as long as no HTTP_PROXY or similar environment variable is set, but will raise that exception if one is.

The Requests maintainers consider the exception to not be a bug in requests (and I agree, although they could document that function better), but the result of passing invalid input for the proxies parameter, which is required to be a dict. See also https://github.com/requests/requests/pull/4636.

On the Bravado side, this code was introduced in 59ef48f572655f6c8f57f1d7878561c16c29c509 by @sjaensch. Fixing it should just be a matter of changing None to {} on line 305, but I haven't tested this. I can confirm that downgrading Bravado to 10.1.0 fixes it for me.

sjaensch:
fixes #401.

### DIFF
--- a/bravado/requests_client.py
+++ b/bravado/requests_client.py
@@ -304,10 +304,10 @@ class RequestsFutureAdapter(FutureAdapter):
         prepared_request = self.session.prepare_request(request)
         settings = self.session.merge_environment_settings(
             prepared_request.url,
-            None,
-            None,
-            self.misc_options['ssl_verify'],
-            self.misc_options['ssl_cert'],
+            proxies={},
+            stream=None,
+            verify=self.misc_options['ssl_verify'],
+            cert=self.misc_options['ssl_cert'],
         )
         response = self.session.send(
             prepared_request,


sjaensch:
The fix has been released as part of bravado 10.2.2.

PR #403, opened 2019-01-03
Initial support for typing

macisamuele:
The goal of this PR is to add typing to the bravado library.
Types are exported via PEP 561 (`py.typed` file).

Notable facts:
 * `bravado.client` and `bravado.docstring_property` are not typed yet (I would prefer doing it in a follow-up)
 * `bravado.compact` has been removed by ensuring `simplejson` dependency for the library.
     This is a _NO OP_ change as `bravado` depends on `bravado-core` which depends on `simplejson` so we were already using `simplejson`
 * `bravado.http_future.HttpFuture.response` logic has been changed a bit to handle the possible case of `self.operation` being `None`.
    To extract the bravado config the original code was running something like `self.operation.swagger_spec.config['bravado']` but if `self.operation` is `None` then an `AttributeError` is thrown. This was never catched by tests but has been identified by `mypy`. In order to avoid changing too many signatures I preferred adding an helper method to extract the configuration object that defaults to the default configurations if `self.operation` is `None`


bravado/fido_client.py
```diff
@@ -33,22 +38,27 @@ def __init__(self, fido_response):
...
         return self._delegate.code
 
     @property
     def text(self):
+        # type: () -> typing.Text
         return self._delegate.body
 
     @property
     def raw_bytes(self):
+        # type: () -> typing.ByteString
```
sjaensch: `byte` should be enough I think?

bravado/fido_client.py
```diff
@@ -66,15 +76,22 @@ def headers(self):
         return self._headers
 
     def json(self, **_):
+        # type: (typing.Any) -> typing.Mapping[str, typing.Any]
         # TODO: pass the kwargs downstream
         return self._delegate.json()
 
 
-class FidoClient(HttpClient):
+class FidoClient(typing.Generic[T], HttpClient):
```
sjaensch: I don't think this is correct generic typing, as the FidoClient class instance is not dependent on this type T - the FidoFutureAdapter is, which is being wrapped by a HttpFuture. Adding the generic typing to HttpFuture is the right move here.

bravado/http_client.py
```diff
@@ -8,7 +14,13 @@ class HttpClient(object):
...
     """
 
-    def request(self, request_params, operation=None, request_config=None):
+    def request(
+        self,
+        request_params,  # type: typing.MutableMapping[str, typing.Any]
+        operation=None,  # type: typing.Optional[Operation]
+        request_config=None,  # type: typing.Optional[RequestConfig]
+    ):
+        # type: (...) -> HttpFuture
```
sjaensch: This should probably be `HttpFuture[T]`.

bravado/http_future.py
```diff
@@ -127,7 +143,21 @@ def __init__(self, future, response_adapter, operation=None,
...
+        # type: () -> BravadoConfig
+        if self.operation:
+            return self.operation.swagger_spec.config['bravado']
+        else:
+            return BravadoConfig.from_config_dict(CONFIG_DEFAULTS)
+
+    def response(
+        self,
+        timeout=None,  # type: typing.Optional[float]
+        fallback_result=SENTINEL,  # type: typing.Optional[typing.Union[typing.Any, typing.Callable[[BaseException], typing.Any]]]  # noqa
```
sjaensch: This is already typed correctly internally - any reason you went with `typing.Any` instead of doing what's done there (i.e. using T)?

bravado/http_future.py
```diff
@@ -200,6 +234,7 @@ def response(self, timeout=None, fallback_result=SENTINEL, exceptions_to_catch=F
         )
 
     def result(self, timeout=None):
+        # type: (typing.Optional[float]) -> typing.Union[T, typing.Tuple[T, IncomingResponse]]
```
sjaensch: It can also be just `IncomingResponse` (the last of the three `return` statements). It's strange that mypy didn't catch this...

bravado/response.py
```diff
@@ -1,24 +1,39 @@
...
     """Bravado response object containing the swagger result as well as response metadata.
 
     :ivar result: Swagger result from the server
     :ivar BravadoResponseMetadata metadata: metadata for this response including HTTP response
     """
 
-    def __init__(self, result, metadata):
+    def __init__(
+        self,
+        result,  # type: R
```
sjaensch: Shouldn't this be T? Also, if you need R then probably the class should be `Generic[R, T]`.

bravado/http_future.py
macisamuele: nice catch

bravado/http_client.py
macisamuele: yup

bravado/fido_client.py
macisamuele: good point
The client itself is not bounded to types (with this I noticed that `bravado.requests_client` is missing typing too)

bravado/http_future.py
macisamuele: I used copy/paste from docstring, but I need to check more as is reasonable to assume that `typing.Any` should be replaced with `T` as we should ensure that the `fallback_result` is of the _same type_ of the endpoint response

bravado/requests_client.py
```diff
@@ -25,12 +29,15 @@ class Authenticator(object):
     """
 
     def __init__(self, host):
+        # type: (str) -> None
         self.host = host
 
     def __repr__(self):
-        return u"%s(%s)" % (self.__class__.__name__, self.host)
+        # type: () -> str
+        return "%s(%s)" % (self.__class__.__name__, self.host)
```
sjaensch: I don't think this is ok, you're changing the return type on Python 2. I assume this was `unicode` on purpose.

bravado/requests_client.py
```diff
@@ -60,13 +68,15 @@ class ApiKeyAuthenticator(Authenticator):
     :param param_in: How to send the API key. Can be 'query' or 'header'.
     """
 
-    def __init__(self, host, api_key, param_name=u'api_key', param_in=u'query'):
+    def __init__(self, host, api_key, param_name='api_key', param_in='query'):
+        # type: (str, str, str, str) -> None
```
sjaensch: Same thing here, I expect us to want `param_name` and `param_in` to be unicode strings and not bytestrings? If we're not 100% sure that this is wrong then we should preserve whatever we have.

bravado/fido_client.py
```diff
@@ -20,6 +20,17 @@
 from bravado.http_future import FutureAdapter
 from bravado.http_future import HttpFuture
 
+if typing.TYPE_CHECKING:
+    class _FidoStub(typing.Protocol):
+        code = None  # type: int
+        body = None  # type: str
+        reason = None  # type: typing.Text
+        headers = None  # type: typing.Mapping[bytes, typing.List[bytes]]
```
sjaensch: This doesn't match the type you're using below, which uses unicode strings instead of bytestrings. I'm wondering why mypy didn't catch this.

bravado/requests_client.py
macisamuele: This change has been made because `__repr__` should be returning a `str`, while we're returning `unicode`.
Now we have two alternatives:
 1) fix the code (as in the PR), the type will be correct but we could have failures related to formatting (NOTE: this is technically possible if `self.host` is `unicode` and it shouldn't )
 2) ignore typing for this specific case

NOTE: In `Authenticator.matches` we ensure that `self.host` is the same of the hostname extracted from the url and `reveal_type(urlparse.urlsplit('').hostname)` tells us that `hostname` will be `str`

Given that I would say that the change should be safe

bravado/requests_client.py
macisamuele: I have to check it more carefully but looks like `param_name` need to have the same type of the values in the headers dictionary.
About `param_in` this is used only internally and is compared with `'headers'`, so the change should be safe.

For `param_in` the fact that is safe does not justify the type change (considering the parameter as part of the public interface) so I'll revert it

bravado/fido_client.py
macisamuele: I would assume that mypy is not helping us out on this conditions because `requests` is not typed.
I have to check whether I can use `typeshed` to have some request typing

bravado/requests_client.py
sjaensch: Sounds good!

bravado/requests_client.py
macisamuele: I've checked the internals of `requests` and looks like `str` refers to `unicode` on python2 ([link|https://github.com/requests/requests/blob/75bdc998e2d430a35d869b2abf1779bd0d34890e/requests/compat.py#L51])

So I'm reverting the `param_name` type change as it is not needed

bravado/fido_client.py
macisamuele: I checked the code and ...
 * `_FidoStub` defines headers to be `Mapping[bytes, List[bytes]]`: this was deduced from [fido|https://github.com/Yelp/fido/blob/10ac95ed5b0e26adfe342147c084f3d0c3908ae0/fido/common.py#L7]
 * `_FidoStub` is used only in `FidoResponseAdapter`.
    `FidoResponseAdapter.heades` need to be conformant to `IncomingResponse.headers` which returns `Mapping[Text, Text]`.
    The code was already decoding the `bytes` key and getting one (the last) value by decoding it. So the end result would be `Mapping[Text, Text]`. This explains the cast of `CaseInsensitiveDict` to `Mapping[Text, Text]`.
    NOTE: the last cast `cast(Mapping[Text, Text], self._headers)` was not needed so I'm removing it

bravado/fido_client.py
```diff
@@ -60,7 +60,7 @@ def text(self):
 
     @property
     def raw_bytes(self):
-        # type: () -> bytes
+        # type: () -> typing.Union[bytes, str]
```
sjaensch: This should always be bytes I thought - can you tell me more about this?

macisamuele: We did check it offline.

The reason for which I've added `Union[bytes, str]` was that:
 * on python2 `self._delegate.body` (for fido) and `self._delegate.content` (for requests) are returning `str`
 * on python3 `self._delegate.body` (for fido) and `self._delegate.content` (for requests) are returning `bytes`

But on python2 `str == bytes` so the `Union` is not needed.

I'm reverting the change

PR #406, opened 2019-02-14
Add ability to cancel a HttpFuture

sjaensch:
Currently there is no way to do so - the workaround would be to call `result(timeout=0)`, but this might raise an exception, and is potentially costly due to all sorts of response processing.

### DIFF
--- a/bravado/fido_client.py
+++ b/bravado/fido_client.py
@@ -218,7 +218,7 @@ class FidoFutureAdapter(FutureAdapter[T]):
         try:
             return self._eventual_result.wait(timeout=timeout)
         except crochet.TimeoutError:
-            self._eventual_result.cancel()
+            self.cancel()
             six.reraise(
                 fido.exceptions.HTTPTimeoutError,
                 fido.exceptions.HTTPTimeoutError(
@@ -228,3 +228,6 @@ class FidoFutureAdapter(FutureAdapter[T]):
                 ),
                 sys.exc_info()[2],
             )
+
+    def cancel(self):
+        self._eventual_result.cancel()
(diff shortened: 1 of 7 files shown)


sjaensch:
I've updated the PR since in the end a backwards-compatible approach might make more sense, but I'm not completely sure yet. What do you guys think? Personally I prefer this since we do print a warning, so it's not like there's no indication that the `cancel()` call is without effect.

lucagiovagnoli:
Mmh, you're adding a cancel() function which was not there before. So people using bravado today, are probably not calling .cancel()
How is this new method backward-breaking?

sjaensch:
Yep it only covers users of external HTTP clients (like bravado-asyncio) starting to use `cancel()` and not realizing that they need to upgrade that package. I thought failing loudly should be fine in that case. But first of all the failure wouldn't necessarily make the fix obvious, nor is it certain that people would hit it during testing (i.e. before they deploy to prod).

That's why I think handling it with a warning might make more sense.

lucagiovagnoli:
On the other hand, a log warning could be lost if the application is configured to ignore warnings. Also who hasn't ignored warnings in logs from time to time.. not sure the warning makes it as obvious as failing loudly. In order to make the fix more informative you might modify the comment: "The FutureAdapter class of the HTTP client does not implement cancel() _in bravado vX.Y.Z_, ignoring the call." That would make it clear it's something related to versioning. That said, I don't have a **very** strong preference for either option, feel free to choose. Ship it!

sjaensch:
Going to go with the warning approach then, running into this crash in production seems like the worse of the two outcomes. Thanks all!

PR #407, opened 2019-02-18
Fix Python 3.5 runtime issues due to typing

sjaensch:
We've run into two issues internally:

1. `typing.TYPE_CHECKING` was added in Python 3.5.2, which means master is currently broken on 3.5.1 and 3.5.0.
2. On Ubuntu xenial with Python 3.5.2, several tests fail with `TypeError: descriptor '__subclasses__' of 'type' object needs an argument`. The code that's failing is a cast we do, which does execute at runtime (but where I'd have thought that it is a no-op). This is the code:
    ```python
    typing.cast(
        typing.List[typing.Union[typing.Type[BaseException], BaseException, typing.Text]],
        sys.exc_info()[:2],
   )

I've manually verified that all tests pass on a xenial box.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -4,6 +4,6 @@ exclude_lines =
     raise AssertionError
     raise NotImplementedError
     if __name__ == .__main__.:
-    if typing.TYPE_CHECKING:
+    if getattr(typing, 'TYPE_CHECKING', False):
     pragma: no cover
     @bottle.(get|post|route)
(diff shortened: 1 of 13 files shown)


macisamuele:
😍 thanks a lot
if tests pass feel free to merge this branch

sjaensch:
I've had to remove additional `cast()` calls, they were causing lots of failing tests with `NotImplementedError: This IncomingResponse type <class 'bravado.fido_client.FidoResponseAdapter'> forgot to implement an attr for headers`. I'm getting uncomfortable with using `cast` at all now...

macisamuele:
Let's make sure that on the Changelog is explicit the fact that `BravadoConfig.from_config_dict` has been removed.
A major release could eventually be avoided by the fact that `BravadoConfig` is an internal structure

PR #425, opened 2019-08-08
Show response text in 5xx errors

KevinColemanInc:
This will show the response body so developers will not need to add a try/except to print body of the response after a failed run.

Related to: https://github.com/Yelp/bravado/pull/32


sjaensch:
Thank you for the contribution @KevinColemanInc ! Could you please merge in master and take a look at the test failures?

KevinColemanInc:
@sjaensch I think the test failures were resolved in the other PR. Merging in master fixed them.

sjaensch:
@KevinColemanInc sorry for the long delay! I'm merging this in as-is, since I think it's quite a nice improvement. One follow-up would be to have a dedicated test for the error formatting, where basically you assert on (and thus show us) what the full formatted exception string would be.

PR #426, opened 2019-08-12
Update client classes - set response & future adapter class as class-level vars

andriis:
Update client classes to have the response and future adapter class as class-level attributes.
This will simplify the custom client class creation - no more need to rewrite the `request` method.

UPD: depends on #427

### DIFF
--- a/CHANGELOG-MASTER.rst
+++ b/CHANGELOG-MASTER.rst
@@ -4,3 +4,6 @@ Changelog-Master
 *This file will contain the Changelog of the master branch.*
 
 *The content will be used to build the Changelog of the new bravado release.*
+
+Update client classes to have response and future adapter class as a class level attributes.
+This will simplify the custom client class creation - no more need to rewrite the `request` method.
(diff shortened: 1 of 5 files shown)


sjaensch:
Hey @andriis, sorry for the long delay in responding to this! It's an interesting idea, what's the use case for having custom future and response adapter classes?

How about instead of using class level attributes we pass instances as optional arguments to `__init__`?

andriis:
Hi @sjaensch, no worries. 

As for the use cases:
- once I had to rewrite the `request` method of `FidoClient` just to specify the custom future adapter with updated timeout log message (with more details);
- `FidoResponseAdapter` is also hardcoded in the `request`, but what if we want to add some logic to the response adapter?

It would be useful to specify these either as the class attributes or in the optional parameters to `__init__`.

I see what you mean, in case of having these in `__init__` we don't need to inherit from `FidoClient` or `RequestsClient` to have the custom adapters in our client class. I'll update the PR to use custom adapters in the `__init__` if you prefer it this way.

andriis:
@sjaensch I opened the new PR #430 to replace this one.

PR #430, opened 2019-08-21
Allow custom future and response adapter classes in clients

andriis:
Allow custom future and response adapter classes in both `RequestsClient` and `FidoClient` through `future_adapter_class` and `response_adapter_class` params respectively.

This PR is opened to replace the #426, in which these adapter classes were proposed as class-level attributes.

### DIFF
--- a/bravado/fido_client.py
+++ b/bravado/fido_client.py
@@ -97,6 +97,21 @@ class FidoClient(HttpClient):
     """Fido (Asynchronous) HTTP client implementation.
     """
 
+    def __init__(
+        self,
+        future_adapter_class=None,  # type: typing.Optional[typing.Type[FidoFutureAdapter]]
+        response_adapter_class=None,  # type: typing.Optional[FidoResponseAdapter]
+    ):
+        # type: (...) -> None
+        """
+        :param future_adapter_class: Custom future adapter class,
+            should be a subclass of :class:`FidoFutureAdapter`
+        :param response_adapter_class: Custom response adapter class,
+            should be a subclass of :class:`FidoResponseAdapter`
+        """
+        self.future_adapter_class = future_adapter_class or FidoFutureAdapter
+        self.response_adapter_class = response_adapter_class or FidoResponseAdapter
+
     def request(
         self,
         request_params,  # type: typing.MutableMapping[str, typing.Any]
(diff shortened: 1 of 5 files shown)


andriis:
@sjaensch what do you think?

bravado/fido_client.py
```diff
@@ -97,6 +97,21 @@ class FidoClient(HttpClient):
     """Fido (Asynchronous) HTTP client implementation.
     """
 
+    def __init__(
+        self,
+        future_adapter_class=None,  # type: typing.Optional[typing.Type[FidoFutureAdapter]]
+        response_adapter_class=None,  # type: typing.Optional[FidoResponseAdapter]
```
sjaensch: This should be `typing.Optional[typing.Type[FidoResponseAdapter]]`.

bravado/requests_client.py
```diff
@@ -122,19 +122,31 @@ class RequestsClient(HttpClient):
...
     """
 
-    def __init__(self, ssl_verify=True, ssl_cert=None):
-        # type: (bool, typing.Any) -> None
+    def __init__(
+        self,
+        ssl_verify=True,  # type: bool
+        ssl_cert=None,  # type:  typing.Any
+        future_adapter_class=None,  # type: typing.Optional[typing.Type[RequestsFutureAdapter]]
+        response_adapter_class=None,  # type: typing.Optional[RequestsResponseAdapter]
```
sjaensch: Same here, we're passing a class and not an instance of the class.

bravado/fido_client.py
```diff
@@ -97,6 +97,21 @@ class FidoClient(HttpClient):
     """Fido (Asynchronous) HTTP client implementation.
     """
 
+    def __init__(
+        self,
+        future_adapter_class=None,  # type: typing.Optional[typing.Type[FidoFutureAdapter]]
```
sjaensch: Any reason you went for `None` as default value instead of `FidoFutureAdapter`?

bravado/fido_client.py
andriis: Oh, you're right, I'll fix it.

bravado/fido_client.py
andriis: @sjaensch I can do it in another PR if needed

PR #431, opened 2019-08-27
Future and response adapter classes as default values

andriis:
This is just a cosmetic change related to #430.
In both `RequestsClient` and `FidoClient` for `future_adapter_class` and `response_adapter_class` values `None` was used. 
This PR is to provide corresponding future adapter and response adapter class values.

### DIFF
--- a/bravado/fido_client.py
+++ b/bravado/fido_client.py
@@ -93,14 +93,54 @@ class FidoResponseAdapter(IncomingResponse):
         return self._delegate.json()
 
 
+class FidoFutureAdapter(FutureAdapter[T]):
+    """
+    This is just a wrapper for an EventualResult object from crochet.
+    It implements the 'result' method which is needed by our HttpFuture to
+    retrieve results.
+    """
+
+    timeout_errors = (fido.exceptions.HTTPTimeoutError,)
+    connection_errors = (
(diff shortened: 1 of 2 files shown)


andriis:
@sjaensch Could you please take a look? It's related to your comment in #430 related to default values for `future_adapter_class` and `response_adapter_class`.

bravado/fido_client.py
```diff
@@ -93,14 +93,54 @@ def json(self, **_):
...
 class FidoClient(HttpClient):
     """Fido (Asynchronous) HTTP client implementation.
     """
 
     def __init__(
         self,
-        future_adapter_class=None,  # type: typing.Optional[typing.Type[FidoFutureAdapter]]
-        response_adapter_class=None,  # type: typing.Optional[typing.Type[FidoResponseAdapter]]
+        future_adapter_class=FidoFutureAdapter,  # type: typing.Optional[typing.Type[FidoFutureAdapter]]
+        response_adapter_class=FidoResponseAdapter,  # type: typing.Optional[typing.Type[FidoResponseAdapter]]
```
sjaensch: Let's remove the `Optional` since `None` is no longer a value we support. Same for the RequestsClient.

andriis: Done

PR #440, opened 2019-11-04
Don't modify request_options['headers'] when constructing request

ngaya-ll:


sjaensch:
Nice catch @ngaya-ll ! Please run the pre-commit hooks on your branch (`tox -e pre-commit`) and commit the result, then this should be good to go.

ngaya-ll:
@sjaensch Fixed the pre-commit issue, all checks are passing now.

PR #441, opened 2019-11-07
Change type of http client requestparams from MutableMapping to Mapping

ngaya-ll:
Request params should not be modified by the http client, as this can lead to unexpected bugs if the caller passes the same parameter dict to multiple requests.


ngaya-ll:
On futher examination, it is not necessary for `request_params` to be immutable at this abstraction level, as the request params are already constructed per-request by the `SwaggerClient`.  I'm going to close this PR.

PR #442, opened 2019-11-11
Remove modindex from docs, configure RTD build

### LINKED ISSUE #285: Read the docs module index broken
Hey, the link to the module index in RTD is broken:

http://bravado.readthedocs.io/en/latest/py-modindex.html

This is linked to from the docs root:

http://bravado.readthedocs.io/en/latest/index.html

How can I view the API docs for Bravado?

sjaensch:
Looks like it's not generated properly anymore on readthedocs. It's not really a problem since we have basically [the same thing](https://bravado.readthedocs.io/en/latest/modules.html) as part of our documentation, but with a better title.

Fixes #285

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,0 +1,14 @@
+# .readthedocs.yml
+# Read the Docs configuration file
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+version: 2
+
+sphinx:
+  configuration: docs/conf.py
+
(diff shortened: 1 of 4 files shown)


sjaensch:
I had some time to look into this, looks like the issue is due to RTD using Python 2.7 for the build, and installing a version of `itertools-more` that is not compatible with it ([example](https://readthedocs.org/api/v2/build/9943438.txt)). Let's see if using a configuration that 

- installs all necessary requirements
- uses Python 3.7

fixes things.

PR #444, opened 2019-11-26
Fix header authentication for host:port type

sanda87:
Hello,
All our apps work on ports and we found out that bravado doesn't support host:port type for Header Authentication. The pull request adds the support.


bravado/requests_client.py
```diff
@@ -46,7 +46,7 @@ def matches(self, url):
         :return: True if matches host, port and scheme, False otherwise.
         """
         split = urlparse.urlsplit(url)
-        return self.host == split.hostname
+        return self.host == split.netloc
```
sjaensch: Hey @sanda87, thanks for the PR and sorry for the late response! 

While this works in your case, we should also either fix the docstring (we're not checking the scheme), or add that in here as well.

Could you also please add a test to make sure it works as expected, and so that there will be no regressions in the future? Maybe `tests/http_client_test.py` is a good fit for that.

sanda87: Hi @sjaensch , 
I added tests for this case and updated docstring in my new commit

PR #447, opened 2020-01-13
Reraise fido HttpTimeoutError without modification

ngaya-ll:
Currently if fido raises a `fido.exceptions.HttpTimeoutError` while performing a request, the original error message will be replaced by the code in `FidoFutureAdapter`.  This can lead to the wrong timeout value being reported in the error message.

There are two cases that can happen.

1. Crochet `EventualResult` does not receive a result within the specified wait timeout, throws a `crochet.TimeoutError`.  In this case the future adapter should wrap the exception with `fido.exceptions.HttpTimeoutError`.  This is the existing functionality.

2. Request fails with `fido.exceptions.HttpTimeoutError`.  In this case the future adapter should reraise the original exception.  Since `fido.exceptions.HttpTimeoutError` is a subclass of `crochet.TimeoutError` we need to explicitly exclude this from case 1.

### DIFF
--- a/bravado/fido_client.py
+++ b/bravado/fido_client.py
@@ -112,10 +112,23 @@ class FidoFutureAdapter(FutureAdapter[T]):
 
     def result(self, timeout=None):
         # type: (typing.Optional[float]) -> T
+
+        # Note: There are two kinds of timeouts that can occur here.
+        #
+        # 1. Fido request fails with a `fido.exceptions.HttpTimeoutError` due
+        # to the configured request timeout. In this case we don't want to
+        # modify the original exception.
+        #
+        # 2. The `EventualResult` is not completed within the specified wait
+        # timeout. In this case we cancel the request and transform the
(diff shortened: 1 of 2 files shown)


bravado/fido_client.py
```diff
@@ -114,6 +114,8 @@ def result(self, timeout=None):
         # type: (typing.Optional[float]) -> T
         try:
             return self._eventual_result.wait(timeout=timeout)
+        except fido.exceptions.HTTPTimeoutError:
+            raise  # Don't modify original exception
```
macisamuele: Please add a comment (like in the PR description) explaining why you are catching the exception and re-raising it without performing actions.

ngaya-ll: Done, let me know if it needs further clarification.

macisamuele: Thanks a lot for doing it.
I've just approved your change.

PR #454, opened 2020-03-19
Disable following of redirects by default

### LINKED ISSUE #449: Bravado should not silently follow redirects
In my API spec, I have an endpoint that returns a 301 redirect. This redirects to another URL on a different server that subsequently returns a 200. However, because bravado is using requests, this redirect never comes back to bravado which gets the 200 (which is not a specified response) and then throws an error.

Relevant section of logs below. Logs without a prefix for where it came from comes from setting `http.client.HTTPConnection.debuglevel = 5`. Comments are inserted for clarity and some addresses are redacted for privacy (as this is on a private network).

```
# here we send the initial request which responds with a 301
[10:11:10.432][bravado.client][DEBUG] getSecureClient({'_request_options': {'headers': {'UserID': '123456'}}, 'storeID': '3888d0f1-7896-4c8e-98c8-4934d5a37aff'})
send: b'GET /api/v2/backup/stores/3888d0f1-7896-4c8e-98c8-4934d5a37aff/secure-client HTTP/1.1\r\nHost: localhost:4000\r\nUser-Agent: python-requests/2.22.0\r\nAccept-Encoding: gzip, d                                        eflate\r\nAccept: */*\r\nConnection: keep-alive\r\nUserID: 123456\r\n\r\n'
reply: 'HTTP/1.1 301 Moved Permanently\r\n'
header: Location: http://<redirected-url>/<somefile>
header: Vary: Origin
header: Date: Mon, 17 Feb 2020 10:11:10 GMT
header: Content-Length: 0
[10:11:10.437][urllib3.connectionpool][DEBUG] http://localhost:4000 "GET /api/v2/backup/stores/3888d0f1-7896-4c8e-98c8-4934d5a37aff/secure-client HTTP/1.1" 301 0

# because bravado doesn't configure anything for redirects
# they are automatically followed by requests/urllib3
[10:11:10.440][urllib3.connectionpool][DEBUG] Starting new HTTP connection (1): web-proxy.sdc.hpecorp.net:8080
send: b'GET http://<redirected-url>/<somefile> HTTP/1.1\r\nHost: 16.26.136.45:8080\r\nUser-Agent: python-requests/2.22.0\r\nAccept-Encoding: gzip, deflate\r\nAccept: */*\r\nConnection: keep-alive\r\nUserID: 123456\r\n\r\n'
reply: 'HTTP/1.1 200 OK\r\n'
header: Date: Mon, 17 Feb 2020 10:12:29 GMT
header: X-Content-Type-Options: nosniff
header: Content-Security-Policy: sandbox; default-src 'none'; img-src 'self'; style-src 'self';
header: X-WebKit-CSP: sandbox; default-src 'none'; img-src 'self'; style-src 'self';
header: X-Content-Security-Policy: sandbox; default-src 'none'; img-src 'self'; style-src 'self';
header: Last-Modified: Mon, 17 Feb 2020 09:49:05 GMT
header: Expires: Mon, 17 Feb 2020 09:49:05 GMT
header: Accept-Ranges: bytes
header: Content-Type: application/octet-stream
header: Content-Length: 17681912
header: Server: Jetty(9.4.z-SNAPSHOT)
header: Connection: Keep-Alive
header: Age: 0
[10:11:10.474][urllib3.connectionpool][DEBUG] http://<proxy>:8080 "GET http://<redirected-url>/<somefile> HTTP/1.1" 200 17681912
  File "/home/matt/repos/baas-cli/.venv/lib/python3.7/site-packages/bravado/http_future.py", line 337, in unmarshal_response
    op=operation,
  File "/home/matt/repos/baas-cli/.venv/lib/python3.7/site-packages/bravado/http_future.py", line 370, in unmarshal_response_inner
    response_spec = get_response_spec(status_code=response.status_code, op=op)
  File "/home/matt/repos/baas-cli/.venv/lib/python3.7/site-packages/bravado_core/response.py", line 160, in get_response_spec
    "status_code or use a `default` response.".format(status_code, op),
bravado_core.exception.MatchingResponseNotFound: Response specification matching http status_code 200 not found for operation Operation(getSecureClient). Either add a response specifi                                        cation for the status_code or use a `default` response.
```

Initially I wondered if it was worth exposing out redirect configuration to the bravado client, but on reflection I don't think bravado should be doing anything in this area. If the API spec says it returns a 301, bravado should not be doing anything extra that might obscure that. I also wondered that if the responses claim they should return a 30X status then redirects should be implicitly disabled, but that creeps into the territory of being a bit too smart or non-obvious.

mattdowdell:
Per #449, redirects should not be silently followed when making requests. This disables that behaviour by default in the synchronous client, and also modifies the responses to no longer consider 3XX HTTP status errors. There's perhaps some nuance to whether all of them are applicable in API responses, but [RFC 7231](https://tools.ietf.org/html/rfc7231) does not appear to consider these to be errors anyway.

I've left in the exception classes for redirects and `make_http_exception` is untouched, largely because I wasn't sure if that was considered part of the public API or not. I've also added an integration test and bench tested using my code that originally found the issue and confirmed that it works as expected for my use case.

Fixes #449.

### DIFF
--- a/bravado/http_future.py
+++ b/bravado/http_future.py
@@ -410,8 +410,8 @@ def raise_on_unexpected(http_response):
 
 def raise_on_expected(http_response):
     # type: (IncomingResponse) -> None
-    """Raise an HTTPError if the response is non-2XX and non-3XX and matches
-    a response in the swagger spec.
+    """Raise an HTTPError if the response is non-2XX and matches a response
+    in the swagger spec.
 
     :param http_response: :class:`bravado_core.response.IncomingResponse`
     :raises: HTTPError
--- a/bravado/http_future.py
+++ b/bravado/http_future.py
@@ -410,13 +410,13 @@ def raise_on_unexpected(http_response):
 
 def raise_on_expected(http_response):
     # type: (IncomingResponse) -> None
-    """Raise an HTTPError if the response is non-2XX and matches a response
-    in the swagger spec.
+    """Raise an HTTPError if the response is non-2XX and non-3XX and matches
+    a response in the swagger spec.
 
     :param http_response: :class:`bravado_core.response.IncomingResponse`
     :raises: HTTPError
     """
-    if not 200 <= http_response.status_code < 300:
+    if not 200 <= http_response.status_code < 400:
         raise make_http_exception(
             response=http_response,
             swagger_result=http_response.swagger_result)
--- a/bravado/http_future.py
+++ b/bravado/http_future.py
@@ -199,7 +199,7 @@ class HttpFuture(typing.Generic[T]):
 
             swagger_result = self._get_swagger_result(incoming_response)
 
-            if self.operation is None and incoming_response.status_code >= 300:
+            if self.operation is None and incoming_response.status_code >= 400:
                 raise make_http_exception(response=incoming_response)
 
             # Trigger fallback_result if the option is set
@@ -276,7 +276,7 @@ class HttpFuture(typing.Generic[T]):
                 return swagger_result, incoming_response
             return swagger_result
 
-        if 200 <= incoming_response.status_code < 300:
+        if 200 <= incoming_response.status_code < 400:
             return incoming_response
 
         raise make_http_exception(response=incoming_response)
--- a/bravado/http_future.py
+++ b/bravado/http_future.py
@@ -199,7 +199,7 @@ class HttpFuture(typing.Generic[T]):
 
             swagger_result = self._get_swagger_result(incoming_response)
 
-            if self.operation is None and incoming_response.status_code >= 400:
+            if self.operation is None and incoming_response.status_code >= 300:
                 raise make_http_exception(response=incoming_response)
 
             # Trigger fallback_result if the option is set
@@ -276,7 +276,7 @@ class HttpFuture(typing.Generic[T]):
                 return swagger_result, incoming_response
             return swagger_result
 
-        if 200 <= incoming_response.status_code < 400:
+        if 200 <= incoming_response.status_code < 300:
             return incoming_response
 
         raise make_http_exception(response=incoming_response)
@@ -416,7 +416,7 @@ def raise_on_expected(http_response):
     :param http_response: :class:`bravado_core.response.IncomingResponse`
     :raises: HTTPError
     """
-    if not 200 <= http_response.status_code < 400:
+    if not 200 <= http_response.status_code < 300:
         raise make_http_exception(
             response=http_response,
             swagger_result=http_response.swagger_result)
--- a/bravado/requests_client.py
+++ b/bravado/requests_client.py
@@ -286,7 +286,6 @@ class RequestsClient(HttpClient):
         ssl_cert=None,  # type:  typing.Any
         future_adapter_class=RequestsFutureAdapter,  # type: typing.Type[RequestsFutureAdapter]
         response_adapter_class=RequestsResponseAdapter,  # type: typing.Type[RequestsResponseAdapter]
-        allow_redirects=False  # type: bool
     ):
         # type: (...) -> None
         """
@@ -306,7 +305,6 @@ class RequestsClient(HttpClient):
         self.ssl_cert = ssl_cert
         self.future_adapter_class = future_adapter_class
         self.response_adapter_class = response_adapter_class
-        self.allow_redirects = allow_redirects
 
     def __hash__(self):
         # type: () -> int
@@ -363,7 +361,6 @@ class RequestsClient(HttpClient):
         misc_options = {
             'ssl_verify': self.ssl_verify,
             'ssl_cert': self.ssl_cert,
-            'allow_redirects': self.allow_redirects,
         }
 
         if 'connect_timeout' in sanitized_params:
@@ -372,6 +369,9 @@ class RequestsClient(HttpClient):
 
         if 'timeout' in sanitized_params:
             misc_options['timeout'] = sanitized_params.pop('timeout')
+
+        misc_options['allow_redirects'] = sanitized_params.pop('allow_redirects', False)
+        logging.warning("misc_options: %r", misc_options)
 
         return sanitized_params, misc_options
 
--- a/bravado/requests_client.py
+++ b/bravado/requests_client.py
@@ -361,6 +361,7 @@ class RequestsClient(HttpClient):
         misc_options = {
             'ssl_verify': self.ssl_verify,
             'ssl_cert': self.ssl_cert,
+            'allow_redirects': sanitized_params.pop('allow_redirects', False),
         }
 
         if 'connect_timeout' in sanitized_params:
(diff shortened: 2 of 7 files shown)


mattdowdell:
@macisamuele Are you able to review?

Also, do you understand the failure in CI for mypy? I don't understand why attributes have gone missing.

bravado/http_future.py
```diff
@@ -199,7 +199,7 @@ def response(
 
             swagger_result = self._get_swagger_result(incoming_response)
 
-            if self.operation is None and incoming_response.status_code >= 300:
+            if self.operation is None and incoming_response.status_code >= 400:
```
macisamuele: I would not change this behavior (at least on this PR) as users might rely on this.

You're right to mention that RFC7231 does not consider them errors.
As this is independent from the specific issue you're trying to solve (following redirects) I would suggest to create an issue for this and eventually publish a new PR.
NOTE: changing this (so 3xx do not raise exception) requires a major version bump

bravado/testing/integration_test.py
```diff
@@ -473,6 +492,16 @@ def test_msgpack_support(self, swagger_http_server):
         assert response.headers['Content-Type'] == APP_MSGPACK
         assert unpackb(response.raw_bytes, encoding='utf-8') == API_RESPONSE
 
+    def test_redirects_are_not_followed(self, swagger_http_server):
```
macisamuele: I would expect a test that ensures that redirects are followed as well.

bravado/testing/integration_test.py
```diff
@@ -473,6 +492,16 @@ def test_msgpack_support(self, swagger_http_server):
...
+        response = self.http_client.request({
+            'method': 'GET',
+            'url': '{server_address}/redirect'.format(server_address=swagger_http_server),
+            'params': {},
+        }).result(timeout=1)
+
+        assert response.status_code == 301
+        assert response.headers['Location'] == '/json'
```
macisamuele: The mypy report is the following
```
bravado/testing/integration_test.py:502: error: Item "Tuple[Any, ...]" of "Union[Any, IncomingResponse, Tuple[Any, IncomingResponse]]" has no attribute "status_code"
bravado/testing/integration_test.py:503: error: Item "Tuple[Any, ...]" of "Union[Any, IncomingResponse, Tuple[Any, IncomingResponse]]" has no attribute "headers"
```

What mypy is trying to say is that response is defined the have following possible types `Any`, `IncomingResponse` or `Tuple[Any, IncomingResponse]`. Not all of them (namely `Tuple[...]`) have `status_code` and `headers` attributes.

In order to fix the mypy report you might try to assert that `response` is of `IncomingResponse` type.
```
assert isinstance(response, IncomingResponse) and response.status_code == 301
```

bravado/testing/integration_test.py
mattdowdell: Done, but only for the `RequestsClient` and it doesn;t look like the fido client supports the ability to follow redirects.

bravado/http_future.py
mattdowdell: Done. Instead I check the response stored in the exception for the relevant information.

bravado/testing/integration_test.py
mattdowdell: That seems to have done the trick. Thanks!

bravado/http_future.py
mattdowdell: Raised #455 to track not raising exceptions when redirects are returned.

bravado/testing/integration_test.py
macisamuele: I think that we should be doing a small effort here to ensure that the two clients behave the same here.
I'll try to look for pointers to help you out here

mattdowdell: That was the conclusion I came to. The logigcal place to add support for following redirects in Fido, would be in [prepare_request_for_twisted](https://github.com/Yelp/bravado/blob/master/bravado/fido_client.py). The retuened value from that in turn goes into [fido.fetch](https://github.com/Yelp/fido/blob/master/fido/fido.py) which attempts to look like the interface for requests, but isn't fully implemented (as demonstrated by the lack of allow_redirects parameter. For the record, I have no clue how the call to fido.fetch works in fido_client.py as it seems to be missing an import.

In summary, the effort required to make fido work with redirects is much more significant that getting requests to disable redirects by default. Given the bug I initially found where following redirects causes bravado to complain about unexpected HTTP statuses, I'd say that support for following redirects at all should probably be removed in the future or it should handle redirects at a higher level, e.g. response returns redirect, so now you should expect the response(s) from the target of the redirect. As not all redirects have a single target according to the spec, that gets even more complicated to handle at a library level, so I'd argue that's an implementation decision for the user of bravado to deal with.

mattdowdell:
@macisamuele Is there anything else you'd like me to address before this gets merged in?

bravado/requests_client.py
```diff
@@ -286,7 +286,6 @@ def __init__(
         ssl_cert=None,  # type:  typing.Any
         future_adapter_class=RequestsFutureAdapter,  # type: typing.Type[RequestsFutureAdapter]
         response_adapter_class=RequestsResponseAdapter,  # type: typing.Type[RequestsResponseAdapter]
-        allow_redirects=False  # type: bool
```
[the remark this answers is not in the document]
mattdowdell: I think you're looking at an old changeset. It was added and then later moved to be a parameter in the `request` method.

bravado/testing/integration_test.py
```diff
@@ -473,6 +494,34 @@ def test_msgpack_support(self, swagger_http_server):
...
+            self.http_client.request({
+                'method': 'GET',
+                'url': '{server_address}/redirect'.format(server_address=swagger_http_server),
+                'params': {},
+            }).result(timeout=1)
+        except HTTPMovedPermanently as exc:
+            assert isinstance(exc.response, IncomingResponse) and exc.response.status_code == 301
+            assert isinstance(exc.response, IncomingResponse) and exc.response.headers['Location'] == '/json'
+        else:
+            pytest.fail("Expected exception was not raised")
```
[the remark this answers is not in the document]
mattdowdell: I hadn't realise that was support by `pytest.raises`.

Changed to that method instead and removed the coverage ignore.

bravado/testing/integration_test.py
```diff
@@ -473,6 +494,34 @@ def test_msgpack_support(self, swagger_http_server):
         assert response.headers['Content-Type'] == APP_MSGPACK
         assert unpackb(response.raw_bytes, encoding='utf-8') == API_RESPONSE
 
+    def test_following_redirects(self, swagger_http_server):
+        # the FidoClient doesn't have a way to turn off redirects being followed
+        # so limit this test to the RequestsClient instead
+        if not isinstance(self.http_client, RequestsClient):
+            pytest.skip('Following redirects is only supported in the RequestsClient')
```
sjaensch: Can we make this a blacklist instead of a whitelist? [bravado-asyncio](https://github.com/sjaensch/bravado-asyncio) does have a way to turn off following redirects, and I'd like to be able to run the test there too.

mattdowdell: Done, although it's somewhat clunky due to how coverage and test dependencies are loaded. If there's better suggestions on how to implement blacklisting without having to load the thing being blacklisted, I'm happy to change it.

sjaensch:
Thanks for your efforts @mattdowdell!

sjaensch:
This has has been released (with a slightly different name for the option, see #457) in bravado 10.6.0.

PR #457, opened 2020-03-30
Rename option from allow_redirect to follow_redirects, add documentation

sjaensch:
I'm renaming the option as it's not so much about allowing redirects, but whether we follow them automatically or not. I'm also documenting the option, and reformatting that table since currently it's too wide, and requires readers to scroll horizontally.

FYI @mattdowdell @macisamuele

### DIFF
--- a/bravado/fido_client.py
+++ b/bravado/fido_client.py
@@ -226,7 +226,7 @@ class FidoClient(HttpClient):
                 'url': string,
                 'timeout': float,  # optional
                 'connect_timeout': float,  # optional
-                'allow_redirects': bool,  # optional
+                'follow_redirects': bool,  # optional
             }
         """
 
--- a/bravado/fido_client.py
+++ b/bravado/fido_client.py
@@ -226,7 +226,6 @@ class FidoClient(HttpClient):
                 'url': string,
                 'timeout': float,  # optional
                 'connect_timeout': float,  # optional
-                'follow_redirects': bool,  # optional
             }
         """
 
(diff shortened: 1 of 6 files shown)


bravado/requests_client.py
```diff
@@ -266,7 +266,7 @@ def result(self, timeout=None):
         response = self.session.send(
             prepared_request,
             timeout=self.build_timeout(timeout),
-            allow_redirects=self.misc_options['allow_redirects'],
+            follow_redirects=self.misc_options['follow_redirects'],
```
mattdowdell: I don't have an objection to it being renamed, but it's worth pointing out that it's called this becaise that's the name for it in `requests` itself. That's also the reason why the tests have started failing. You can fix them by changing the parameter name here back to `allow_redirects` if you want to pursue the rename.

docs/source/configuration.rst
```diff
@@ -113,26 +113,43 @@ Configuration can also be applied on a per-request basis by passing in
...
+                                             | regardless of whether the request
+                                             | succeeded.
+                                             | Mainly useful for manual and automated
+                                             | testing.
+*follow_redirects*        boolean   False    | Whether redirects returned by the server
+                                             | are followed, or returned as-is.
+                                             | **Note:** Currently, the fido HTTP client
+                                             | does not support returning redirect
+                                             | responses directly, and will always follow
+                                             | redirects.
```
mattdowdell: That's not correct - the FidoClient will never follow redirects.

bravado/fido_client.py
```diff
@@ -226,7 +226,7 @@ def prepare_request_for_twisted(request_params):
                 'url': string,
                 'timeout': float,  # optional
                 'connect_timeout': float,  # optional
-                'allow_redirects': bool,  # optional
+                'follow_redirects': bool,  # optional
```
mattdowdell: I think this was incorrectly added in my PR per the lack of other changes in this file. I couldn't get this to work, but I don't believe I later removed it from the documentation.

PR #468, opened 2020-10-05
Make sure we use IPv4 for integration tests

sjaensch:
They fail for bravado-asyncio on Python3.7+ otherwise due to aiohttp trying to connect using IPv6 first, but bottle not listening on `::1`.


analogue (approved):
🚢

PR #476, opened 2021-04-05
Migrate to github actions

analogue:
CORESERV-10770


analogue:
`tox -e mypy` was already broken and I couldn't fix it in 5 mins so leaving that for another day.

PR #484, opened 2023-02-14
Use standard library mock when possible

carlwgeorge:
Mock has been in the standard library since 3.3. Use it when available, falling back to the mock from PyPI if necessary.


carlwgeorge:
@benbariteau Would you mind reviewing this so the work doesn't accidentally get duplicated like what happened in #485?

PR #485, opened 2023-02-14
Prefer getfullargspec instead of deprecated getargspec

carlwgeorge:
`inspect.getargspec` has been deprecated since Python 3.0.  It is even removed in Python 3.11.  `inspect.getfullargspec` is the replacement.


carlwgeorge:
Looks like this was fixed in 3324ff5817ade8746454b1f09a65eed4540e915f.

PR #498, opened 2025-05-06
Drop monotonic dependency

carlwgeorge:
Since Python 3.3, monotonic.monotonic is an alias for the standard library time.monotonic.


carlwgeorge:
@benbariteau Would you mind reviewing this so the work doesn't accidentally get duplicated like what happened in #485?

PR #500, opened 2025-05-07
drop old pythons

benbariteau:
All of our builds were in EOL pythons. This raises the minimum version to 3.8 and adds builds for 3.10 and 3.11 as well.

Some additional changes here:
- Bump CI base ubuntu to something also not-EOL
- mypy error due to compatibility logic in `bravado/exception.py` was simplest to fix by just dropping the python 2 logic
- pytest needed an upgrade to support python 3.10+
- `inspect.getargspec` was removed in python 3.11, so I had to fix a test

This repo could use a lot of TLC, but I want to unblock https://github.com/Yelp/bravado/pull/499

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,19 +8,22 @@ jobs:
       fail-fast: false
       matrix:
         python:
-          - py38
-          - py310
-          - py311
+          - env: py38
+            version: '3.8'
+          - env: py310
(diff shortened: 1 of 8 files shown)


jtwang (approved):
Not going to pretend I could catch breakages, but if tests are happy then so am I. Expunging Python 2 everywhere is good!

PR #504, opened 2026-05-06
Deprecate use_msgpack in favor of Accept header

ltrottier-yelp:
# Summary

  - Adds a DeprecationWarning when use_msgpack is used in _request_options
  - Guides users to use headers={"Accept": "application/msgpack"} directly instead

# Context

  Part of a multi-repo change to support msgpack content negotiation via standard HTTP headers:
  1. bravado-core https://github.com/Yelp/bravado-core/pull/419 — marshal/unmarshal based on Content-Type header
  2. This PR — deprecate the now-redundant use_msgpack shorthand

use_msgpack only ever set Accept: application/msgpack. With the new standard header approach, users can set both Content-Type and Accept directly, making use_msgpack redundant.

# Test plan
  - Existing test updated to verify DeprecationWarning is emitted
  - All construct_request tests passing


bravado/client.py
```diff
@@ -322,6 +323,12 @@ def construct_request(operation, request_options, **op_kwargs):
     }
     # Adds Accept header to request for msgpack response if specified
     if request_options.get('use_msgpack', False):
+        warnings.warn(
+            "use_msgpack is deprecated. Use headers={'Accept': 'application/msgpack'} "
+            "in _request_options instead.",
```
ymilki: Rather than specify the exact way to do it.
```suggestion
            "use_msgpack is deprecated. Set the request headers "
            "{'Accept': 'application/msgpack'}",
```

ymilki (changes requested):

